Issue #11, opened 2020-09-02
prevent overwriting entity variables


basnijholt:
I just accidentally did
```
light.living_room_lights = "off"
```
and now my lights in Home Assistant became unavailable.

Maybe it an idea to now allow overwriting entities? 

The same goes for `state`, `service`, `task`, etc.

craigbarratt:
I don't know why they become unavailable.  Is it because the attributes get cleared when you do this?

This seems related to issue #9.

craigbarratt:
With commit 712692f, state variable setting now preserves attributes.

fleXible:
While playing around I came across a different problem regarding overwriting variables.
I seems, that declared variables take presedence over provided variables/objects.

Sample code:
````python
@time_trigger
def precende_problem():
    log.info(f"set_sensor_daytime_startup: running")
    event = None
    event.fire("raises Exception")
````
> precende_problem: running
Exception in <jupyter_0.precende_problem> line 35:
        event.fire("raises Exception")
        ^
AttributeError: 'NoneType' object has no attribute 'fire'

I came across this, when doing a for loop and naming the variable "state" without thinking too much about it.
It wouldn't be too much of an issue, if variables would be scoped, but as they are not, weird things can happen if you don't watch out.

**Comment by craigbarratt** [comment]
Variable scoping should be the same as Python, and I agree it would be a bug if it wasn't.  If you create a local variable it will overwrite a previously defined function or object.

The example looks correct to me.  This behaves exactly the same in Python and pyscript:
```python
bytes.fromhex("1234")
bytes = None
bytes.fromhex("5678")
```
The first call works, and the second one fails with an error:
```
AttributeError: 'NoneType' object has no attribute 'fromhex'
```
Here's another example that also works the same in Python and pyscript:
```python
import math
print(math.sqrt(2))
math = None
print(math.sqrt(2))
```
I agree there is a subtle difference in pyscript: yes, `event` or `state` don't exist as an object.  But if it does (ie: you set them to something), that takes precedence over the `event.fire` or `state.set` functions, which is the same behavior as regular Python.

fleXible:
What you say makes sense, I just never happen to stumble upon it before. Issue can be closed.

Issue #9, opened 2020-09-02
no way to prevent state_set from removes existing attributes


fleXible:
Hi,

using `state.set("domain.entity", "new state")` removes existing attributes. It is possible to supply an attribute dict, but there was no way of collecting all existing attributes from a state. `state.get()` only supported pulling a single attribute by name.

Also `state_get()` would not return the full State object, but only the state as string or the attribute value.

As a possible way to fix this, I did some changes:
- Add `states` object, known from [template extensions](https://www.home-assistant.io/docs/configuration/templating/) to be the new way of simple state retrievals
- Add new method `get_new` that returns the actual State object and remapped `state.get()` to it
- Add new method `set_new` that preserves existing attributes by default, but also allowes them to be updated or removed

The PR will follow soon.

craigbarratt:
Good suggestions- it would be good to have a clean way of providing those features.  This is related to issues #11 and #12.

craigbarratt:
I just made a commit 712692f that makes state setting preserve the attributes.  `state.set()` can optionally update or replace attributes.  I also added `state.get_attr()` to get a dict of all attributes for a state variable.

craigbarratt:
This is released in 0.21.

If you can suggest additional features or changes, please re-open this issue or create a new one.

fleXible:
@craigbarratt 
In my PR I also fancied the idea to have access to the full `State`object, maybe to access any of the other available properties like `last_changed`or `context`.

And I suggested using the `states`objects from HAs template engine. Even though the way I implemented it, was quite ugly, still it lets you iterate over all states or thos from a specific domain and gives access to the full `State` object.
I liked it, because it works the same way as in templates, making it easier for people to adopt `pyscript`.

What do you think?

PR #13, opened 2020-09-02
add function entity_ids that returns all entities

### LINKED ISSUE #12: add function 'entity_ids'
Right now I cannot figure out how to get a list of all defined entities.

Maybe there could be a function
```python
def entity_ids(domain: str = None) -> List[str]:
    """Returns a list of all entity_ids in a domain. If None, all entity_ids are returned."""
```

basnijholt:
Closes #12

### DIFF
--- a/README.md
+++ b/README.md
@@ -265,6 +265,9 @@ In cases where you need to compute the name of the state variable dynamically, o
 set the state attributes, you can use the built-in functions `state.get(name)` and
 `state.set(name, value, attr=None)`; see below.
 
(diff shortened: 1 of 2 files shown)


basnijholt:
I've tested this and it works as intended :tada:

basnijholt:
@craigbarratt I rebased and tested that it still works.

craigbarratt:
Ok, thanks.  We might change this when we address issues #9 and #11.

craigbarratt:
I replaced `entity_ids()` with `state.names()` in commit 7c087e7.

PR #22, opened 2020-09-12
Feat/cron syntax

fleXible:
Had a good run today ;)

### DIFF
--- a/custom_components/pyscript/manifest.json
+++ b/custom_components/pyscript/manifest.json
@@ -4,7 +4,7 @@
   "config_flow": false,
   "documentation": "https://github.com/custom-components/pyscript",
   "issue_tracker": "https://github.com/custom-components/pyscript/issues",
-  "requirements": [],
+  "requirements": ["croniter==0.3.34"],
   "ssdp": [],
   "zeroconf": [],
   "homekit": {},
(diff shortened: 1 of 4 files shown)


craigbarratt:
Thanks - `croniter` provides more features than what I wrote.

craigbarratt:
By the way, I've been learning about pre-commit hooks and I've added some based on the ones in HASS (black/flake8/isort etc).  However, I'm not sure how to create `.git/hooks/pre-commit` which is required to run the pre-commit hooks - it's not part of the repository.  Does there need to be a setup script that creates it after someone clones the repository?

fleXible:
> Thanks - `croniter` provides more features than what I wrote.
I liked the ease of use. 2 or 3 lines of code and that's all I needed!

> Does there need to be a setup script that creates it after someone clones the repository?
Yes, there's a python module called [pre-commit](https://pre-commit.com/) for it.
It is really easy to use. But I don't really like pre-commit hooks that much, because I use the IDE **PyCharm**, which more or less does all the important parts directly while I'm coding. And I usually only commit, once I'm happy with it. Waiting for some hooks at that point just hurts me.

I prefer to have a good Push-Integration like **GitHub Actions**, for those rare cases when I missed to run the final tests. But that's just personal preference.

PR #31, opened 2020-10-09
Add config flow support

raman325:
This change makes it possible to setup pyscript from the UI so a restart is not required (but `configuration.yaml` is still supported so there is no breaking change). I also updated the code to create the `<config/path>/pyscript` folder if it doesn't exist - not sure if there was an intentional reason why you required the user to create the folder but it seemed like an easy improvement (and otherwise, the user would have to create the pyscript folder *before* setting the integration up in the UI).

### DIFF
--- a/custom_components/pyscript/config_flow.py
+++ b/custom_components/pyscript/config_flow.py
@@ -1,10 +1,13 @@
 """Config flow for Vizio."""
-import copy
+import voluptuous as vol
 
 from homeassistant import config_entries
 
-from . import PYSCRIPT_SCHEMA
 from .const import CONF_ALLOW_ALL_IMPORTS, DOMAIN
+
+PYSCRIPT_SCHEMA = vol.Schema(
+    {vol.Optional(CONF_ALLOW_ALL_IMPORTS, default=False): bool}, extra=vol.ALLOW_EXTRA,
+)
 
 
 class PyscriptConfigFlow(config_entries.ConfigFlow, domain=DOMAIN):
@@ -16,9 +19,10 @@ class PyscriptConfigFlow(config_entries.ConfigFlow, domain=DOMAIN):
     async def async_step_user(self, user_input):
         """Handle a flow initialized by the user."""
         if user_input is not None:
-            if await self.async_set_unique_id(unique_id=DOMAIN, raise_on_progress=True):
-                self.async_abort(reason="single_instance_allowed")
+            if len(self.hass.config_entries.async_entries(DOMAIN)) > 0:
+                return self.async_abort(reason="single_instance_allowed")
 
+            await self.async_set_unique_id(DOMAIN)
             return self.async_create_entry(title=DOMAIN, data=user_input)
 
         return self.async_show_form(step_id="user", data_schema=PYSCRIPT_SCHEMA)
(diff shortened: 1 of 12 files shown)


dlashua:
If this PR is accepted, it paves the way for, in my opinion, a very useful feature.

Currently, we can use `state.set()` in pyscript. However, this has several caveats that go with it: 

1) the entities it "creates" are not restored, 
2) existing entities with the same entity_id are blindly overwritten,
3) these entities cannot be managed (renamed, etc) in the Home Assistant UI

Using Config Flow and the appropriate Home Assistant Entity Objects, pyscript could support `pyscript.binary_sensor.create()` and `pyscript.binary_sensor.update()` as well similar methods for `sensor` to allow for pyscript automations to actually create entities/devices in the Home Assistant way.

[This Code](https://github.com/zachowj/hass-node-red) which allows creating of such entities/devices over websocket commands for use with Node-Red should provide a good example of how this is done.

custom_components/pyscript/config_flow.py
```diff
@@ -0,0 +1,48 @@
+"""Config flow for Vizio."""
```
dlashua, line 1: Change to "pyscript"

raman325: that's embarrassing, thanks!

craigbarratt:
Thanks for this PR!  It looks good.  Yes, there was no good reason to have the user create the pyscript directory - I saw that `python_scripts` did it that way.  Thanks for adding tests too.

craigbarratt:
I ran `pylint` and got the following warnings:
```
************* Module pyscript.config_flow
custom_components/pyscript/config_flow.py:19:4: W0222: Signature differs from overridden 'async_step_user' method (signature-differs)
custom_components/pyscript/config_flow.py:19:4: W0222: Signature differs from overridden 'async_step_user' method (signature-differs)
```

craigbarratt:
@dlashua reported and I've confirmed this issue.

This PR has broken the initial loading of config settings.  It used to be that the `config_entry` argument to `async_setup_entry()` contained the current yaml configuration, starting with a top-level entry for `pyscript`.  The `load_files()` function extract's pyscript's `apps` config with:
```
apps_config = config.get(DOMAIN, {}).get("apps", None)
```
That's now `None` for two reasons:
* `config_entry.data` starts below `"pyscript"`, not at `"pyscript"`
* `config_entry.data` is some old setting - it's not the current yaml configuration

On `reload` it works correctly since it explicitly loads the latest yaml config with:
```
config = await async_process_component_config(hass, conf, integration)
```
That might be an expensive operation, so I'm not sure whether it's ok to do that on startup.

raman325:
Thanks for the heads up @craigbarratt clearly missed some things. I actually think the way that reload is working would not work with config entries. We shouldn't be loading data from the config, we should be loading the data from the `config_entry`. I've already fixed the issue I think, I just need to figure out how to resolve the reload before I submit a new PR

craigbarratt:
I actually don't understand how config flow works, and how the configuration is stored.  Can you mix yaml with new-style config?

The reason this is important is that just before your PR, I committed a change that allows other config parameters (below `pyscript`) to be available for pyscript apps and scripts.

So I guess we need a way to harmonize new-style and yaml config, for both startup and reload.

raman325:
You mean can you add YAML config info in addition to setting up the config through the UI? I don't believe so unless you had a separate YAML file for the additional config items that your integration read and loaded in.

The way config flow works is that you define a schema for a config, and then the UI shows it to the user to fill out based on the data types -> once the user submits, the backend sees a dictionary as you would expect from a YAML config and stores that to `data` in the config_entry. To handle YAML configs, there is an "import" flow that reads in the YAML and then passes that dictionary to the backend call that creates the entry, essentially mimicking a user doing the same.

craigbarratt:
So perhaps we should have a separate yaml configuration file for pyscript apps?  @dlashua - what do you think?

raman325:
There is one other option - We can make it so that if a user sets up through the UI, we can import just the apps section of the config on a reload and update the config entry. I would have to think about how that should work, but it is an option.

PR #35, opened 2020-10-11
Fix pylint error from PR #31

raman325:
Should fix https://github.com/custom-components/pyscript/pull/31#issuecomment-706630554


raman325:
As an aside, it might be a good idea to add `pylint` as a pre-commit hook

craigbarratt:
Thanks - it's clean now.  `pylint` isn't part of the commit/PR tests; seems a bit onerous to add it.  But I do check it from time-to-time.

PR #41, opened 2020-10-14
Fix config flow strings

raman325:
As discussed in the other PR thread, originally I had attempted to include references to HA's common strings, but that doesn't work for custom integrations so I'm providing the text here


craigbarratt:
Thanks!

PR #49, opened 2020-10-18
Simplify setting of attributes

swazrgb:
* Support omitting the value for `state.set('domain.value', some_attr='abc')`

* Support setting attributes like `domain.value.some_attr = 'abc'`

* Support setting attributes like `state.set_attr('domain.value.attr', 'abc')`

### DIFF
--- a/custom_components/pyscript/eval.py
+++ b/custom_components/pyscript/eval.py
@@ -1167,9 +1167,15 @@ class AstEval:
                 return
             if not isinstance(var_name, str):
                 raise NotImplementedError(f"unknown lhs type {lhs} (got {var_name}) in assign")
-            if var_name.find(".") >= 0:
+            dot_count = var_name.count(".")
+            if dot_count == 1:
                 State.set(var_name, val)
                 return
+            if dot_count == 2:
+                State.set_attr(var_name, val)
+                return
+            if dot_count > 0:
+                raise NotImplementedError("variable names may contain at most 2 dots")
             if self.curr_func and var_name in self.curr_func.global_names:
                 self.global_sym_table[var_name] = val
                 return
--- a/custom_components/pyscript/state.py
+++ b/custom_components/pyscript/state.py
@@ -98,12 +98,19 @@ class State:
         return notify_vars
 
     @classmethod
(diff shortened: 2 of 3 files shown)


craigbarratt:
This looks good - thanks for submitting.

Could you also include updates to tests for the new features, and docs to reflect these changes?

swazrgb:
I've updated the docs.

I assume I should be looking at `test_function.py` and update/add tests there? I'm trying to understand the result of `literal_eval(await wait_until_done(notify_q))`, but am a bit lost.

I was thinking about using the following pyscript to test:

```python
pyscript.test_attr.a = False
pyscript.test_attr.b = False

@service
def test_set_attr(key, value):
    state.set(f"pyscript.test_attr.{key}", value)

@service
def test_attr_a(value):
    pyscript.test_attr.a = value

@service
def test_attr_b(value):
    pyscript.test_attr.a = value

@state_trigger("True or pyscript.test_attr")
def test_attr():
    pyscript.test_attr = "on" if any([pyscript.test1.a, pyscript.test1.b]) else "off"
```

craigbarratt:
It's probably sufficient to add expressions and results to `tests/test_unit_eval.py`.

The first test function in that file executes all the tests in the long list.  The list has pairs of [source_code, expected_result].

So you could do things like:
```
    ["state.set_attr('x.y.attr1', 10); x.y.attr1", 10],
```
You should test both directions (ie, setting the variable directly and checking the value with `state.get_attr` etc, and also testing the new form of `state.set()` without a value.

The second function checks for exceptions.  So you could add tests that raise specific exceptions, and confirm you get the correct error.  Generally when I write those, I first put in an empty error message so the test fails, and I can see what error is generated.  So long as that looks correct, I then paste that in as the expected error message.

craigbarratt:
You'll need to rebase given the changes to master.

Also, are you planning to add some tests per the discussion above?

craigbarratt:
@swazrgb - do you plan to update this PR?  Or do you want me to manually merge the changes?

craigbarratt:
Manually added these changes, with a few tweaks, and added tests.

PR #53, opened 2020-10-22
Unload scripts and unsubscribe from event listeners during entry unload

raman325:
Implements changes discussed in #42

### DIFF
--- a/custom_components/pyscript/__init__.py
+++ b/custom_components/pyscript/__init__.py
@@ -211,6 +211,9 @@ async def async_unload_entry(hass, config_entry):
     # Unload scripts
     await unload_scripts()
 
+    # tell reaper task to exit (after other tasks are cancelled)
+    await Function.reaper_stop()
+
     # Unsubscribe from listeners
     for unsub_listener in hass.data[DOMAIN][UNSUB_LISTENERS]:
         unsub_listener()
--- a/custom_components/pyscript/__init__.py
+++ b/custom_components/pyscript/__init__.py
@@ -23,10 +23,12 @@ from homeassistant.loader import bind_hass
 from .const import (
     CONF_ALLOW_ALL_IMPORTS,
     CONF_HASS_IS_GLOBAL,
+    CONFIG_ENTRY,
     DOMAIN,
     FOLDER,
     LOGGER_PATH,
     SERVICE_JUPYTER_KERNEL_START,
+    UNSUB_LISTENERS,
 )
 from .eval import AstEval
 from .event import Event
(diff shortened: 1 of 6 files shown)


craigbarratt:
This definitely look like the right approach.  I'm busy for at least the next day, so I'll check it out further then.

raman325:
Just to add, I added a call to `Function.reaper_stop()` in `async_unload_entry` to prevent the following error from occurring due to the test:
```
run_coro: got exception Traceback (most recent call last):
  File "/Users/raman/projects/pyscript/custom_components/pyscript/function.py", line 234, in run_coro
    await coro
RuntimeError: coroutine ignored GeneratorExit
```

not sure if this call belongs in `async_unload_entry` or in `unload_scripts` so I used my best judgement

craigbarratt:
Thanks.  I'll test it after I merge.

raman325:
Sounds good. One thing I realized that may be worth pointing out is that performing a reload through the Integrations UI will not reload the configuration.yaml information, it will just reload the integration and the data that already exists in the config entry. If it is desirable to have both reloads (service and integrations UI) I can look at how to achieve that.

raman325:
Doing what I suggested above would be relatively easy I think, we just need to copy this logic: https://github.com/custom-components/pyscript/blob/master/custom_components/pyscript/__init__.py#L103-L114 and add it here: https://github.com/custom-components/pyscript/blob/master/custom_components/pyscript/__init__.py#L90

This also begs the question, if we are trying to mimick the reload service, should the service be retired and instead direct users to use the Integrations UI? The advantage of having a service is that it can be called from an automation, whereas a UI reload can't

craigbarratt:
Great - that's basically what I did in 0d4c5a6.  UI reload and the reload service should be the same now. I confirmed yaml config updates work in both cases, and task cleanup seems correct now too.

Yes, I'd like to keep the service so you can call it from Jupyter or an automation, as you point out.

raman325:
Nice! 👍

raman325:
By the way, I'm sorry I didn't realize the true complexity of trying to add config flow support, I naively thought it would be simpler. But the end result is pretty cool!

dlashua:
I would love to see the ability to add an "app" (even if I have to plug the YAML into a form input) from the UI. I, personally, won't use it much. But, I know that would make it a lot easier for some other users.

craigbarratt:
@raman325 - no problem at all.  I'm a beginner on HASS so it's helpful for me to understand more of its features. Thanks for your diligence and quick follow-up on questions or tweaks to the code.  Your additions turned out nicely.

Yes, if apps in pyscript become popular, we could figure out a way to add config flow support for them as you suggest.  Perhaps we would have some way for a pyscript app to define its schema (eg, setting a global variable like an integration does), and that could be used in config flow and for error checking.  But I don't think it's worth doing that yet.

raman325:
👍 I wouldn't call myself an HA expert but I'm a relatively frequent contributor so happy to help where I can.

PR #59, opened 2020-10-28
Add context for state_trigger and event_trigger

### LINKED ISSUE #50: Feature Request: service call contexts
Right now, when a pyscript performs a service call, the LogBook has an entry like this:

```
HallDown Occupied turned off by service input_boolean.turn_off
```

However, when Home Assistant Automations perform an action, the LogBook looks like this:

```
 halldown_overhead turned off by halldown_occupied_off 
```

It does this through a `context` object:

```python
        service_task = self._hass.async_create_task(
            self._hass.services.async_call(
                domain,
                service,
                service_data,
                blocking=True,
                context=self._context,
                limit=limit,
            )
        )
```

pyscript can create one of these with something like:

```python
from homeassistant.core import Context

# context comes from the state_trigger, time_trigger, etc
parent_id = None if context is None else context.id
new_context = Context(parent_id=parent_id)
```

I can't find a lot of documentation on how to create context and name the pieces so they show up well in LogBook, but I think this is a start anyway. I think even this small bit of code would make the logbook show that, for instance, the "binary_sensor.dark" in my @state_trigger is what called input_boolean.turn_on. And, even this, is better than what we have now.

The final bit would be to get LogBook to show that binary_sensor.dark activated my pyscript called "halldown_at_dark", and that "halldown_at_dark" is what called input_boolean.turn_on.

dlashua:
This partially resolves #50.

Test with the following:

```python
@state_trigger(['input_boolean.test_1'])
    log.info(data)


@event_trigger('PYSCRIPT_TEST')
    log.info(data)
```

`context` should be visible in both cases of `data`.

### DIFF
--- a/custom_components/pyscript/__init__.py
+++ b/custom_components/pyscript/__init__.py
@@ -187,6 +187,7 @@ async def async_setup_entry(hass, config_entry):
             "var_name": var_name,
             "value": new_val,
             "old_value": old_val,
+            "context": event.context,
         }
         await State.update(new_vars, func_args)
 
--- a/custom_components/pyscript/event.py
+++ b/custom_components/pyscript/event.py
@@ -39,6 +39,7 @@ class Event:
         func_args = {
             "trigger_type": "event",
             "event_type": event.event_type,
+            "context": event.context,
         }
         func_args.update(event.data)
         await cls.update(event.event_type, func_args)


dlashua:
Superseded by https://github.com/custom-components/pyscript/commit/d0745ea200dc966aa089cf6d7e5638ee7f94d631.

PR #60, opened 2020-10-28
event.fire and state.set send context if provided

dlashua:
Test with:

```python
from homeassistant.core import Context


@state_trigger(['input_boolean.test_1'])
    log.info(data)
    new_context = Context(parent_id=data['context'].id)
    state.set('binary_sensor.test', 'on', context=new_context)

@state_trigger(['binary_sensor.test'])
def test_context_sent(**data):
    log.info(data)

@event_trigger('PYSCRIPT_TEST')
    log.info(data)
    new_context = Context(parent_id=data['context'].id)
    event.fire('PYSCRIPT_TEST2', context=new_context)

@event_trigger('PYSCRIPT_TEST2')
def test_context_sent_event(**data):
    log.info(data)
```

data logged by `test_context_sent_event` and `test_context_sent` should show `parent_id` of the received context.

### DIFF
--- a/custom_components/pyscript/function.py
+++ b/custom_components/pyscript/function.py
@@ -117,7 +117,12 @@ class Function:
     @classmethod
     async def event_fire(cls, event_type, **kwargs):
         """Implement event.fire()."""
-        cls.hass.bus.async_fire(event_type, kwargs)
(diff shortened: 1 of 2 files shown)


craigbarratt:
Looks great - thanks!

PR #61, opened 2020-10-29
allow context on short form service calls

dlashua:
allow short form service calls (i.e. `input_boolean.turn_on(entity_id='whatever')`) to send context


dlashua:
Superseded by #62

PR #62, opened 2020-10-29
Logbook Functionality

dlashua:
There might be a better way to do this instead of using `Function` like I have. However, it works.

A few things to note.

* The event fired before the function runs can be named anything we want. I chose "pyscript_running" and then realized Home Assistant uses "automation_triggered". This can be renamed by changing it in both `trigger.py` and `logbook.py`.

* While I have confirmed that `state.set()` is sending the correct context to the best of my knowledge, it seems that states set in this manner don't show up in the log book at all. The state still "sets" as expected, but there is no logbook entry. Therefore, all of the "context" code can be removed from `state.set()` if desired.

* I've left the original context as part of the trigger function kwargs. This way, pyscript users can still look at `user_id` to determine who did something. Unfortunately, determining if **pyscript** did something would mean climbing the context tree, which is tedious.

* I've also left context as an optional kwarg to `service.call` and `event.fire`. If a users passes a context, it'll use that. If they don't and we have one from the `pyscript_running` event we fired, it'll use that. If they want to disable the logbook functionality for some reason, they can pass `context=Context()`.

* The logbook attributes actions to entity_ids. And entity ids can only have one "." between the domain and name. So I'm sending, essentially, `__name__`, with "." converted to "_". This means the entity_ids are a bit ugly (ex. `pyscript.file_myfile_myfunction`). This is set in `trigger.py` and can be changed to anything we want as long as it fits the rules for entity ids.

I've tested events, service_calls, time triggers, and state.set with this code:

```python
@state_trigger(["input_boolean.test_1"])
    log.info("test_state_trigger")
    input_boolean.toggle(entity_id="input_boolean.test_2")
    state.set("binary_sensor.test", "on")


@event_trigger("PYSCRIPT_TEST")
    log.info("test_event_trigger")
    input_boolean.toggle(entity_id="input_boolean.test_3")


@state_trigger(["binary_sensor.test"])
def report_binary_sensor(**data):
    log.info(f"binary_sensor trigger with {data}")


@time_trigger("startup")
def test_time_trigger(**data):
    task.sleep(3)
    log.info(f"test_time_trigger {data}")
    input_boolean.toggle(entity_id="input_boolean.test_4")
```

It results in this logbook:

logbook

### DIFF
--- a/custom_components/pyscript/function.py
+++ b/custom_components/pyscript/function.py
@@ -125,10 +125,8 @@ class Function:
         if "context" in kwargs and isinstance(kwargs["context"], Context):
             context = kwargs["context"]
             del kwargs["context"]
-        elif curr_task in cls.task2context:
-            context = cls.task2context[curr_task]
         else:
-            context = Context()
+            context = cls.task2context.get(curr_task, None)
 
         cls.hass.bus.async_fire(event_type, kwargs, context=context)
 
@@ -195,10 +193,8 @@ class Function:
         if "context" in kwargs and isinstance(kwargs["context"], Context):
             context = kwargs["context"]
             del kwargs["context"]
-        elif curr_task in cls.task2context:
-            context = cls.task2context[curr_task]
         else:
-            context = Context()
+            context = cls.task2context.get(curr_task, None)
 
         await cls.hass.services.async_call(domain, name, kwargs, context=context)
 
@@ -261,10 +257,8 @@ class Function:
             if "context" in kwargs and isinstance(kwargs["context"], Context):
                 context = kwargs["context"]
                 del kwargs["context"]
-            elif curr_task in cls.task2context:
-                context = cls.task2context[curr_task]
             else:
-                context = Context()
+                context = cls.task2context.get(curr_task, None)
 
             await cls.hass.services.async_call(domain, service, kwargs, context=context)
 
--- a/custom_components/pyscript/function.py
+++ b/custom_components/pyscript/function.py
@@ -239,7 +239,12 @@ class Function:
             return None
 
         async def service_call(*args, **kwargs):
-            await cls.hass.services.async_call(domain, service, kwargs)
+            if "context" in kwargs and isinstance(kwargs["context"], Context):
+                context = kwargs["context"]
+                del kwargs["context"]
+                await cls.hass.services.async_call(domain, service, kwargs, False, context)
+            else:
+                await cls.hass.services.async_call(domain, service, kwargs)
 
         return service_call
 
--- a/custom_components/pyscript/function.py
+++ b/custom_components/pyscript/function.py
@@ -134,6 +134,7 @@ class Function:
 
     @classmethod
     def store_hass_context(cls, hass_context):
+        """Store a context against the running task."""
         curr_task = asyncio.current_task()
         cls.task2context[curr_task] = hass_context
 
--- a/custom_components/pyscript/function.py
+++ b/custom_components/pyscript/function.py
@@ -27,6 +27,10 @@ class Function:
     unique_name2task = {}
 
     #
+    # Mappings of task id to hass contexts
+    task2context = {}
+
+    #
     # Set of tasks that are running
     #
     our_tasks = set()
@@ -117,12 +121,21 @@ class Function:
     @classmethod
     async def event_fire(cls, event_type, **kwargs):
         """Implement event.fire()."""
+        curr_task = asyncio.current_task()
         if "context" in kwargs and isinstance(kwargs["context"], Context):
             context = kwargs["context"]
             del kwargs["context"]
-            cls.hass.bus.async_fire(event_type, kwargs, context=context)
+        elif curr_task in cls.task2context:
+            context = cls.task2context[curr_task]
         else:
-            cls.hass.bus.async_fire(event_type, kwargs)
+            context = Context()
+
+        cls.hass.bus.async_fire(event_type, kwargs, context=context)
+
+    @classmethod
+    def store_hass_context(cls, hass_context):
+        curr_task = asyncio.current_task()
+        cls.task2context[curr_task] = hass_context
 
     @classmethod
     def task_unique_factory(cls, ctx):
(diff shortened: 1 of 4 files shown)


dlashua:
Also, this changes the way #61 works. So that PR is not needed if this one is merged. However, I'm leaving that open for now in case this PR needs more work since that change is useful.

dlashua:
I lied about the `state.set()` not showing up in the logbook. It wasn't showing up because I was setting it to the same value for state. Once I adjusted my code to actually CHANGE the state, this code logs exactly as expected.

custom_components/pyscript/function.py
```diff
@@ -117,12 +121,21 @@ async def async_sleep(cls, duration):
...
+        curr_task = asyncio.current_task()
         if "context" in kwargs and isinstance(kwargs["context"], Context):
             context = kwargs["context"]
             del kwargs["context"]
-            cls.hass.bus.async_fire(event_type, kwargs, context=context)
+        elif curr_task in cls.task2context:
+            context = cls.task2context[curr_task]
         else:
-            cls.hass.bus.async_fire(event_type, kwargs)
+            context = Context()
```
craigbarratt, line 131: The last 4 lines could just be:
```
else:
    context = cls.task2context.get(curr_task, None)
```

custom_components/pyscript/function.py
```diff
@@ -177,12 +190,16 @@ def service_has_service(cls, domain, name):
...
+        curr_task = asyncio.current_task()
         if "context" in kwargs and isinstance(kwargs["context"], Context):
             context = kwargs["context"]
             del kwargs["context"]
-            await cls.hass.services.async_call(domain, name, kwargs, context=context)
+        elif curr_task in cls.task2context:
+            context = cls.task2context[curr_task]
         else:
-            await cls.hass.services.async_call(domain, name, kwargs)
+            context = Context()
```
craigbarratt, line 201: The last 4 lines could just be:
```
else:
    context = cls.task2context.get(curr_task, None)
```

custom_components/pyscript/function.py
```diff
@@ -239,12 +256,16 @@ def get(cls, name):
...
+            curr_task = asyncio.current_task()
             if "context" in kwargs and isinstance(kwargs["context"], Context):
                 context = kwargs["context"]
                 del kwargs["context"]
-                await cls.hass.services.async_call(domain, service, kwargs, False, context)
+            elif curr_task in cls.task2context:
+                context = cls.task2context[curr_task]
             else:
-                await cls.hass.services.async_call(domain, service, kwargs)
+                context = Context()
```
craigbarratt, line 267: The last 4 lines could just be:
```
else:
    context = cls.task2context.get(curr_task, None)
```

custom_components/pyscript/logbook.py
```diff
@@ -0,0 +1,45 @@
...
+        ev_trigger_type = func_args.get("trigger_type", "unknown")
+        if ev_trigger_type == "event":
+            ev_source = f"event {func_args.get('event_type', 'unknown event')}"
+        elif ev_trigger_type == "state":
+            ev_source = f"state change {func_args.get('var_name', 'unknown entity')} == {func_args.get('value', 'unknown value')}"
+        elif ev_trigger_type == "time":
+            ev_trigger_time = func_args.get("trigger_time", "unknown")
+            if ev_trigger_time is None:
+                ev_trigger_time = "startup"
+            ev_source = f"time {ev_trigger_time}"
```
craigbarratt, line 32: Would it make sense (and be possible) to have a user-specified message that could override the defaults here?  That would allow the pyscript developer to optionally format their own messages that could be more informative based on the situation.

Related to that, is there a mechanism for pyscript to create its own logbook entries?

I'm happy to leave both of these issues for a future PR, rather than burden this one with supporting this too.  It is probably best to use the new features in this PR for a while before deciding how best to address user-defined logbook entries, or user-defined logbook messages for the builtin trigger, service and state logbook entries.

custom_components/pyscript/state.py
```diff
@@ -117,10 +118,14 @@ async def set(cls, var_name, value, new_attributes=None, **kwargs):
...
 
-        context = None
+        curr_task = asyncio.current_task()
         if "context" in kwargs and isinstance(kwargs["context"], Context):
             context = kwargs["context"]
             del kwargs["context"]
+        elif curr_task in Function.task2context:
+            context = Function.task2context[curr_task]
+        else:
+            context = Context()
```
craigbarratt, line 128: The last 4 lines could just be:
```
else:
    context = cls.task2context.get(curr_task, None)
```

custom_components/pyscript/trigger.py
```diff
@@ -826,14 +827,31 @@ async def trigger_watch(self):
                     )
                     continue
 
+                # Create new HASS Context with incoming as parent
+                if "context" in func_args and isinstance(func_args["context"], Context):
+                    hass_context = Context(parent_id=func_args["context"].id)
+                else:
+                    hass_context = Context()
```
craigbarratt, line 834: can this just be `hass_context = None`?

craigbarratt (commented):
Looks great!  Here are a few minor comments...

custom_components/pyscript/trigger.py
dlashua: It makes little difference. If you pass it None, then when the call is processed by HASS, it'll make a Context() for it. If you generate a Context() then it will skip that step. Same number of cycles either way. I changed it to use "None" in all cases where I was sending an empty Context()

custom_components/pyscript/logbook.py
dlashua: To my knowledge, the only way to create a logbook entry is to call a service or fire an event. When you call a service, the logbook determines what that looks like on its own. When you fire an event, the "logbook.py" functionality here is the only thing I know of. Beyond that, the logbook.log service is available:

In pyscript...
```python
logbook.log(name="Kitchen", message="is being used", entity_id="domain.name", domain="light")
```

I will play with this and see if context is carried and the future service calls attributed to the logbook entry when done like this.

Allowing a user specified message would be awesome. If logbook.log allows that, then that's perfect. If not, I don't see any way to do it using the event.fire mechanism since we have to fire the event before the function even runs. Maybe with some decorator magic?

dlashua: In testing, `logbook.log` can generate any logbook message. However, if a subsequent service call is tied to it by context, then, it appears in the logbook as "test_2 turned on by service logbook.log" which is not useful at all.

So, users can use `logbook.log` to be more verbose about what their scripts are doing if desired, but the event firing with the correct context (and subsequent contexts) is the only way to tie them together.

Additionally, I tried firing a properly crafted event without the corresponding function in `logbook.py`. With this, the event does not show up in the logbook, and the subsequent state changes are attributed only to the service call.

dlashua:
I'd like it if the text used to indicate what is doing the work in the logbook was clearer. The logbook is weird, though, so it comes from two places, as illustrated here:

logbook2_LI

As a result, it makes sense to keep the entity_id and the name the same (but the entity_id would have "pyscript." at the beginning of it, which the logbook does not display). 

I think maybe what should be displayed is something like "pyscript function test_script in file/app mytest". This should give the user a pretty clear indication of where to look for what caused this if they are debugging.

PR #68, opened 2020-10-31
Allow users to create a requirements.txt to install missing packages

### LINKED ISSUE #66: Feature Request: Support python package installation somehow
I love that `pyscript` lets you import any package to use in automations, but I am somewhat limited in the packages I can use because I run HomeAssistant in a Docker container, and I would have to either manually install the packages each time I rebuilt the container or create my own `Dockerfile` that did the same. It would be great if there was a mechanism through `pyscript` to install packages if they are not found. I am not sure what the best way to achieve this would be, there are two problems to consider:
1) How to determine what packages need to be installed and their versions? HomeAssistant achieves this through the use of `manifest.json` which lists the dependent package and version.
2) How to install the package? You can probably use the code here as a reference https://github.com/home-assistant/core/blob/dev/homeassistant/requirements.py#L105

Open questions:
1) How would `pyscript` deal with version mismatches between what HomeAssistant expects and what `pyscript` wants? My take is that we should always defer to HomeAssistant requirements to prevent issues with the core application

Thanks for the consideration!

raman325:
As I mentioned in #66, it would be nice to be able to have `pyscript` install packages to be used in imports in cases where the user doesn't have the ability to install the packages themselves (in my case, I am using a Docker container for HomeAssistant so it would be cumbersome to install these requirements on every container rebuild).

I went ahead and implemented the functionality in this PR for your consideration. The way it works is that users can add a `requirements.txt` file to `pyscript/`. For any packages that are already installed, the version will be ignored so as not to mess with HomeAssistant functionality, and for any packages that aren't already installed, the package will be installed via a Home Assistant function. Since I am leveraging the Home Assistant function, the actual work being done within pyscript is fairly minimal.

This activity to install requirements will be done on initialization and on reload.

Resolves #66

### DIFF
--- a/custom_components/pyscript/__init__.py
+++ b/custom_components/pyscript/__init__.py
@@ -147,7 +147,7 @@ async def async_setup_entry(hass, config_entry):
 
     State.set_pyscript_config(config_entry.data)
 
-    await install_requirements(hass)
+    await hass.async_add_executor_job(install_requirements, hass)
     await load_scripts(hass, config_entry.data)
 
     async def reload_scripts_handler(call):
@@ -165,7 +165,7 @@ async def async_setup_entry(hass, config_entry):
 
         await unload_scripts(global_ctx_only=global_ctx_only)
 
-        await install_requirements(hass)
+        await hass.async_add_executor_job(install_requirements, hass)
         await load_scripts(hass, config_entry.data, global_ctx_only=global_ctx_only)
 
         start_global_contexts(global_ctx_only=global_ctx_only)
@@ -292,7 +292,7 @@ async def install_requirements(hass):
                     if requirement_installed_version in requirement:
                         _LOGGER.debug("`%s` already found", requirement.project_name)
                     else:
-                        _LOGGER.debug(
+                        _LOGGER.warning(
                             (
                                 "`%s` already found but found version `%s` does not"
                                 " match requirement. Keeping found version."
@@ -308,12 +308,14 @@ async def install_requirements(hass):
                     # Not valid requirements line so it can be skipped
                     _LOGGER.debug("Ignoring `%s` because it is not a valid package", pkg)
             if requirements_to_install:
-                _LOGGER.info("Installing the following packages: %s", ", ".join(requirements_to_install))
+                _LOGGER.info(
+                    "Installing the following packages from %s: %s",
+                    requirements_path,
+                    ", ".join(requirements_to_install),
+                )
                 await async_process_requirements(hass, DOMAIN, requirements_to_install)
             else:
-                _LOGGER.info("All requirements are already available.")
-    else:
-        _LOGGER.info("No requirements.txt found so nothing to install.")
+                _LOGGER.debug("All packages in %s are already available", requirements_path)
 
 
 @bind_hass
--- a/custom_components/pyscript/__init__.py
+++ b/custom_components/pyscript/__init__.py
@@ -269,53 +269,52 @@ async def unload_scripts(global_ctx_only=None, unload_all=False):
 @bind_hass
 async def install_requirements(hass):
     """Install missing requirements from requirements.txt."""
-    requirements_path = os.path.join(hass.config.path(FOLDER), "requirements.txt")
+    for root in ("", "apps/*", "modules/*"):
+        for requirements_path in glob.glob(os.path.join(hass.config.path(FOLDER), root, "requirements.txt")):
+            with open(requirements_path, "r") as requirements_file:
+                requirements_to_install = []
+                for pkg in requirements_file.readlines():
+                    # Remove inline comments which are accepted by pip but not by Home
+                    # Assistant's installation method.
+                    # https://rosettacode.org/wiki/Strip_comments_from_a_string#Python
+                    i = pkg.find("#")
+                    if i >= 0:
+                        pkg = pkg[:i]
+                    pkg = pkg.strip()
 
-    if os.path.exists(requirements_path):
-        with open(requirements_path, "r") as requirements_file:
-            requirements_to_install = []
-            for pkg in requirements_file.readlines():
-                # Remove inline comments which are accepted by pip but not by Home
-                # Assistant's installation method.
-                # https://rosettacode.org/wiki/Strip_comments_from_a_string#Python
-                i = pkg.find("#")
-                if i >= 0:
-                    pkg = pkg[:i]
-                pkg = pkg.strip()
+                    try:
+                        # Attempt to get version of package. Do nothing if it's found since
+                        # we want to use the version that's already installed to be safe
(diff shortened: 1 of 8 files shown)


raman325:
As I was thinking through this, there is a "slicker" way to do this, which would be to add the logic to install the module during import time (https://github.com/custom-components/pyscript/blob/master/custom_components/pyscript/eval.py#L843 and https://github.com/custom-components/pyscript/blob/master/custom_components/pyscript/eval.py#L876). The advantage of this approach is that the user doesn't have to add a `requirements.txt` or do anything really to modify their code, we could just check that the module is available and install it if not. The disadvantage is that there would be no way to pin to a version. I personally prioritize stability over convenience, which suggests that my original approach is better, but I am presenting that option in case you are against the idea of adding another file to manage and value convenience over stability (stability coming from the ability to pin a version).

dlashua:
I think I prefer the requirements.txt approach. Not only does it allow pinning a version, but, if we're installing a module, it's best to be explicit about that. This approach will provide the least amount of surprise. Especially if someone requests a module that takes forever to build (like numpy) and doesn't have a prebuilt binary available.

custom_components/pyscript/__init__.py
```diff
@@ -250,6 +266,55 @@ async def unload_scripts(global_ctx_only=None, unload_all=False):
...
+                try:
+                    # Attempt to get version of package. Do nothing if it's found since
+                    # we want to use the version that's already installed to be safe
+                    requirement = pkg_resources.Requirement.parse(pkg)
+                    requirement_installed_version = version(requirement.project_name)
+
+                    if requirement_installed_version in requirement:
+                        _LOGGER.debug("`%s` already found", requirement.project_name)
+                    else:
+                        _LOGGER.debug(
```
craigbarratt, line 295: This should probably be a `LOGGER.warning` level message, since it's important for the user to know if there is a version mismatch that is being ignored.

custom_components/pyscript/__init__.py
```diff
@@ -250,6 +266,55 @@ async def unload_scripts(global_ctx_only=None, unload_all=False):
         await GlobalContextMgr.delete(global_ctx_name)
 
 
+@bind_hass
+async def install_requirements(hass):
+    """Install missing requirements from requirements.txt."""
+    requirements_path = os.path.join(hass.config.path(FOLDER), "requirements.txt")
```
craigbarratt, line 272: See the discussion in the PR comments - it would be great to also check `requirements.txt` files in app and module subdirectories too.

custom_components/pyscript/__init__.py
```diff
@@ -250,6 +266,55 @@ async def unload_scripts(global_ctx_only=None, unload_all=False):
         await GlobalContextMgr.delete(global_ctx_name)
 
 
+@bind_hass
+async def install_requirements(hass):
```
craigbarratt, line 270: This function does I/O and could be quite slow installing requirements.  So it has to run outside the main event loop.  It should be run in another thread via `hass.async_add_executor_job`.

custom_components/pyscript/__init__.py
```diff
@@ -250,6 +266,55 @@ async def unload_scripts(global_ctx_only=None, unload_all=False):
...
+                except ValueError:
+                    # Not valid requirements line so it can be skipped
+                    _LOGGER.debug("Ignoring `%s` because it is not a valid package", pkg)
+            if requirements_to_install:
+                _LOGGER.info("Installing the following packages: %s", ",".join(requirements_to_install))
+                await async_process_requirements(hass, DOMAIN, requirements_to_install)
+            else:
+                _LOGGER.info("All requirements are already available.")
+    else:
+        _LOGGER.info("No requirements.txt found so nothing to install.")
```
craigbarratt, line 316: Probably ok to remove this log message

custom_components/pyscript/__init__.py
```diff
@@ -250,6 +266,55 @@ async def unload_scripts(global_ctx_only=None, unload_all=False):
...
+                    _LOGGER.debug("%s not found, adding it to package installation list", pkg)
+                    requirements_to_install.append(pkg)
+                except ValueError:
+                    # Not valid requirements line so it can be skipped
+                    _LOGGER.debug("Ignoring `%s` because it is not a valid package", pkg)
+            if requirements_to_install:
+                _LOGGER.info("Installing the following packages: %s", ",".join(requirements_to_install))
+                await async_process_requirements(hass, DOMAIN, requirements_to_install)
+            else:
+                _LOGGER.info("All requirements are already available.")
```
craigbarratt, line 314: This could be made a `debug` level message, and/or made more explicit - eg,  "All packages in requirements.txt files are already available"

custom_components/pyscript/__init__.py
raman325: I can make that change, but are there any concerns that we will load files and attempt to run them/import packages before requirements are installed?

raman325: nevermind, we would still await it. I will submit a new PR shortly

craigbarratt:
I definitely like this approach using `requirements.txt`.  Implicitly installing packages at import time is undesirable for several reasons - Python doesn't do it, it could be confusing, and it doesn't allow version pinning as @dlashua noted.

I added a few comments to the PR, which is looking good.

One request is to support multiple `requirements.txt` files in the following directories:
* `pyscript/requirements.txt`: top-level already supported
* `pyscript/apps/APP/requirements.txt`: if an app uses the package (directory) form, it can optionally specify its own requirements.  If you use the simple module form `pyscript/apps/APP.py` then you can't specify requirements.
* `pyscript/modules/MODULE/requirements.txt` ditto for modules.

The motivation for the last two is it allows apps or modules to be self-contained and they will just do the right thing when installed.

custom_components/pyscript/__init__.py
raman325: `add_executor_job` expects a synchronous function so it never gets awaited. So back to my original point, don't we *want* to block the loop so we pause execution of the rest of the routine until the packages get installed? If not, imports will fail

craigbarratt: Yes, we definitely want to block until it completes.  However, we can't have the blocking I/O in the async function.  So just the synchronous blocking parts should go in a function called by `add_executor_job`.  See `load_files` for how it does that - it does all the I/O in single synchronous function.  In a similar way, you could have a single function that does all the file globs and file reading, and it could return a combined str, or list of the file contents, or better yet a hash indexed by file name (useful for error message reporting).  The async function could then goes ahead and parses each file's str contents and does the necessary installs.

raman325: I see what you mean, and this may also help me fix the test. Stay tuned.

craigbarratt:
This all looks great.  Merging.  Thanks!

PR #69, opened 2020-10-31
Refactor logic for processing requirements.txt so all I/O operations happen in a single function

raman325:
`installed_version` also does IO from #68, this moves it to an executor job. Thanks @bdraco for catching this!


craigbarratt:
Good catch.  However, I'd rather not run a new thread inside the loop for every package.

Perhaps most of the code can move into `load_all_requirement_lines`, and it just returns a list of packages that need to be installed.  Then `install_requirements` just calls `hass.async_add_executor_job` and `async_process_requirements`.  That way we only run one thread at startup or reload, not 1 + the number of lines in all the requirements files.

craigbarratt:
Nice!  This looks great, and tests look good too.

Issue #70, opened 2020-11-01
Open Discussion: How to handle package installations for duplicate but differing requirements


raman325:
One limitation with the current implementation for package installation is that two different `requirements.txt` can list the same package with different version requirements that may not be compatible (e.g. `pkg==0.0.1` and `pkg==1.0.0`), and the package will be installed multiple times in the order that it was parsed which could lead to potential issues down the line. What should the right behavior be? There are a couple of decisions to make:
1. Do we try to resolve this at all or do we let it happen with a caveat in the documentation?
2. If we try to resolve this, we can make a best effort to find a version that matches all of the requirements, but if we can't, does the newer version or the older version take precedence?

The step to find a version that matches all of the requirements would normally be achieved by passing a constraints file to `pip`, but we are using Home Assistant's installation mechanism which uses its own constraints file, so that's not a valid option for us. We would have to essentially implement our own version of this, which we may be able to use `pkg_resources` to do, I just haven't had a chance to look into it yet, and I wanted to get your thoughts @craigbarratt before I spend any time on this

raman325:
One other thing to note - this will only be an issue in a scenario where the package doesn't already exist and we process two requirements files with the same package for installation in the same cycle. Once the package has been installed, future conflicts will be discarded based on our current logic, so maybe this isn't worth trying to implement a solution for.

dlashua:
I'm not sure exactly what information you have available to make decisions but, if possible...

If Home Assistant (or any other custom_component, etc) requires a certain version of a package, that should be the version that gets installed. We should never change something Home Assistant has already done. This will help ensure that Home Assistant doesn't break as a result of package installation.

If Home Assistant doesn't care about a certain package, then we should install the highest version number requested by any pyscript for that package and, if it break, it breaks. The user can sort it out. We should be willing to downgrade a package, though. So, if I install version 5 of something and it breaks something else that asks for version 3 and then I realize I can do just fine with version 3 so I change the requirements, on the next pass it should see that the highest requested version is now 3 and attempt to downgrade to that.

The docs should advise users to pin to the least specific version that will still work in order to provide flexibility.

raman325:
The code already prevents pyscript from overwriting an existing package -> HomeAssistant will always install over a version if the existing version mismatches what it needs, whether that's for core functionality or custom_components.

[...]

PR #72, opened 2020-11-02
For package installations, require version pinning, handle version mismatches with higher version winning, and store installed packages to handle downgrades

raman325:
As mentioned in #70 the original implementation of installing packages from `requirements.txt` had some limitations, primarily around version mismatches. The previous code didn't do anything to address the case where different pyscript `requirements.txt`'s had different version requirements for the same package, it would simply tell HA to install them in the order they were found, resulting in the last version found being installed.

This change does several things:
1. Tracks packages installed by `pyscript` - This is necessary to handle scenarios where users want to downgrade a package they installed through `pyscript`. We want to let them downgrade through `pyscript` but not if it was installed via another source.
2. Requires version pinning - This makes conflict resolution simpler because it makes it possible to do the next item.
3. Always defer to the highest version defined in `requirements.txt`. The current implementation can easily change to match different logic depending on what we want to do here. My opinion is we either defer to the highest version or the lowest when there is a conflict.
4. Install any package version that is either not already installed or was installed by `pyscript` (and this handles if it was originally installed by `pyscript` but another actor came in after the fact and changed the version) - this would allow users to downgrade packages as needed but would protect Home Assistant in case it takes over package management for a package installation that `pyscript` used to manage

There is complexity introduced by this change, but I tried to be comprehensive in my tests to account for that. I also moved the package logic into a separate module to keep `__init__.py` clean and to make it easier to write tests, so the number of lines changed looks bigger than it actually is.


raman325:
I thought about this a bit, and the reason for version pinning is really to avoid handling all the different operator types. I can probably update this so that either you don't specify a version at all, or if you do, you *have* to do it by pinning. I think that would be a better user experience, so I will look into that.

raman325:
Created a separate PR for handling unpinned versions as mentioned above. This will allow the maintainer to decide which, if any, features should be supported.

custom_components/pyscript/requirements.py
```diff
@@ -0,0 +1,223 @@
...
+                if (
+                    len(requirement.specs) != 1
+                    or len(requirement.specs[0]) != 2
+                    or requirement.specs[0][0] != "=="
+                ):
+                    _LOGGER.error(
+                        (
+                            "Ignoring invalid requirement `%s` specified in `%s`, version must "
+                            "be pinned using the format `pkg==version`"
+                        ),
```
craigbarratt, line 89: All the LOG message inner backquotes should normal single quotes; I didn't notice these before.

tests/test_requirements.py
```diff
@@ -0,0 +1,178 @@
...
+            "my-package-name-alternate": {
+                ATTR_SOURCES: [f"{PYSCRIPT_FOLDER}/requirements.txt"],
+                ATTR_VERSION: "2.0.1",
+                ATTR_INSTALLED_VERSION: "1.2.1",
+            },
+        }
+        await install_requirements(hass, entry, PYSCRIPT_FOLDER)
+        await hass.async_block_till_done()
+        assert not ha_install_requirements.called
+        assert "WARNING" in caplog.text and "installed outside of pyscript" in caplog.text
```
craigbarratt, line 87: This assert could be combined into a single longer string, including the name of the package, so that it's more specific.

When I write these caplog asserts I first put in something that fails.  Based on the log output, I find the string in the log, confirm it's what I expect, and then paste that into the caplog assert statement.

craigbarratt:
Both this PR and #78 look good.  I do like the approach discussed in #70: packages installed outside take precedence, and within pyscript the highest version wins, with warnings about unsatisfied requirements.

I added a couple of very minor comments.  I'm happy to merge this PR, then have you rebase #78.  Or more simply I can just merge #78.  So if you want you could just resolve the comments in #78 

One other request (for this PR or #78) is to add some info to the docs so the priority and conflict resolution is explained.  And also add a bullet to `docs/new_features.rst`, or update the existing one (that text will become the release notes on the next version).

Thanks for the good ideas and implementation here - this is another excellent addition.

raman325:
Thanks for the feedback, I will incorporate the changes - I didn't add documentation because I wasn't sure where to put it, but I am happy to take a stab at that. Given that you like both this and #78, I will close this and we can focus on the single PR

Issue #73, opened 2020-11-02
state_trigger firing although value and old_value are the same.


kmorris81:
This state_trigger keeps tripping even though the actual state doesn't appear to have changed.  I will be at home and regularly/randomly get notifications on my phone from this function:

```
@state_trigger("device_tracker.my_iphone == 'home'")
def someone_arrived(**kwargs):
    log.info(f"Someone arrived, disarming system: {kwargs}")
    alarm_control_panel.alarm_disarm(entity_id="alarm_control_panel.ha_alarm")
    notify.mobile_app_my_iphone(title="Status", message="Someone arrived, disarming system.")
```

The logging above provides this output:

`INFO (MainThread) [custom_components.pyscript.file.example.someone_arrived] Someone arrived, disarming system: {'trigger_type': 'state', 'var_name': 'device_tracker.my_iphone', 'value': 'home', 'old_value': 'home', 'context': Context(user_id=None, parent_id=None, id='xxxxxxxxxxxxxxxxxx')}`

As you can see, both `value` and `old_value` are `home`; so I'm not sure exactly why this is happening.

My interim-fix is to manually check `value` and `old_value` inside the function and return if they are equal.

craigbarratt:
I just confirmed an attribute change will cause HASS to send a new state changed event, even if the value doesn't change. That seems unexpected, but we'll have to live with it.  I guess I need to figure out the exact cases where a state changed event can happen and make sure it is clearly documented.

So your workaround is the right one.  You could also equivalently do it in the `@state_trigger`:
```
@state_trigger("device_tracker.my_iphone == 'home' and device_tracker.my_iphone.old != 'home'")
```

kmorris81:
Thanks for the quick and informational response.

None of my other state_triggers seem to exhibit this behavior.  Perhaps it's unique to the mobile_app tracking and HASS companion app?

The fallout isn't a huge deal and neither is the workaround; I just wanted to verify that I wasn't doing something stupid or see if this was a known caveat to the pyscript implementation.

Thanks again!

craigbarratt:
It would be great to confirm that it is in fact an attribute change causing the state change notification.  Assuming you still have the `if` statement inside the function, before that can you log that state's attributes?  This should do it:
```
log.info(f"device_tracker.my_iphone attributes = {state.get_attr('device_tracker.my_iphone')}")
```

dlashua:
[...]

Issue #30, opened 2020-10-05
task.sleep doesn't sleep long enough (in a class?)


dlashua:
This script:
```
registered_triggers = []

class SleepTest:

    def __init__(self, name, seconds):

        @time_trigger('startup')
        def inner_trigger():
            log.error(f"{name}: starting")

            log.error(f"{name}: sleeping {seconds}")
            task.sleep(seconds)
            log.error(f"{name}: done sleeping")

        registered_triggers.append(inner_trigger)

task.sleep(1)
SleepTest('short', 5)
SleepTest('long', 30)
```

This log:
```
2020-10-05 15:43:16 ERROR (MainThread) [custom_components.pyscript.file.sleeptest.inner_trigger] short: starting
2020-10-05 15:43:16 ERROR (MainThread) [custom_components.pyscript.file.sleeptest.inner_trigger] short: sleeping 5
2020-10-05 15:43:16 ERROR (MainThread) [custom_components.pyscript.file.sleeptest.inner_trigger] long: starting
2020-10-05 15:43:16 ERROR (MainThread) [custom_components.pyscript.file.sleeptest.inner_trigger] long: sleeping 30
2020-10-05 15:43:21 ERROR (MainThread) [custom_components.pyscript.file.sleeptest.inner_trigger] short: done sleeping
2020-10-05 15:43:26 ERROR (MainThread) [custom_components.pyscript.file.sleeptest.inner_trigger] long: done sleeping
```

Notice the timestamps, long did not sleep long enough.

dlashua:
Note: The task.sleep(1) is just to make sure this runs later than everything else. I have quite a few pyscripts running that generate output. This sleep just makes sure it's at the end of the startup log spam.

craigbarratt:
I tried that in both Jupyter and a `sleeptest.py` file, and it works correctly for me.  Here's the output:
```
2020-10-05 14:03:41 ERROR (MainThread) [custom_components.pyscript.file.sleeptest.inner_trigger] short: starting
2020-10-05 14:03:41 ERROR (MainThread) [custom_components.pyscript.file.sleeptest.inner_trigger] short: sleeping 5
2020-10-05 14:03:41 ERROR (MainThread) [custom_components.pyscript.file.sleeptest.inner_trigger] long: starting
2020-10-05 14:03:41 ERROR (MainThread) [custom_components.pyscript.file.sleeptest.inner_trigger] long: sleeping 30
[...]

PR #76, opened 2020-11-03
WIP: detect old attribute values

dlashua:
This addresses #73 .

This may not be correct, but limited testing shows it works for me.

Old attribute values are available to state trigger as `binary_sensor.test.old.attribute_name`.

I've also added `change` to `func_args` to indicate that a change was detected. 

@state_trigger takes kwarg `only_on_change` (default: True) 

I will continue to update this PR as I test.

### DIFF
--- a/custom_components/pyscript/__init__.py
+++ b/custom_components/pyscript/__init__.py
@@ -232,7 +232,7 @@ async def async_setup_entry(hass, config_entry):
 
         for attribute in event.data["new_state"].attributes:
             new_val = event.data["new_state"].attributes[attribute]
-            if "old_state" in event.data and attribute in event.data["old_state"].attributes:
+            if "old_state" in event.data and event.data['old_state'] is not None and attribute in event.data["old_state"].attributes:
                 old_val = event.data["old_state"].attributes[attribute]
             else:
                 old_val = None
--- a/custom_components/pyscript/__init__.py
+++ b/custom_components/pyscript/__init__.py
@@ -211,7 +211,7 @@ async def async_setup_entry(hass, config_entry):
             new_val = event.data["new_state"].state
         old_val = event.data["old_state"].state if event.data["old_state"] else None
 
-        if old_val != new_val:
+        if old_val == new_val:
             val_change = False
         else:
             val_change = True
@@ -237,7 +237,7 @@ async def async_setup_entry(hass, config_entry):
             else:
                 old_val = None
 
-            if old_val != new_val:
+            if old_val == new_val:
                 val_change = False
             else:
                 val_change = True
--- a/custom_components/pyscript/__init__.py
+++ b/custom_components/pyscript/__init__.py
@@ -210,15 +210,48 @@ async def async_setup_entry(hass, config_entry):
         else:
             new_val = event.data["new_state"].state
         old_val = event.data["old_state"].state if event.data["old_state"] else None
+
+        if old_val != new_val:
+            val_change = False
+        else:
+            val_change = True
+
         new_vars = {var_name: new_val, f"{var_name}.old": old_val}
         func_args = {
             "trigger_type": "state",
             "var_name": var_name,
             "value": new_val,
             "old_value": old_val,
+            "change": val_change,
             "context": event.context,
         }
         await State.update(new_vars, func_args)
+
+        if "new_state" not in event.data or event.data["new_state"] is None:
+            return
+
+        for attribute in event.data["new_state"].attributes:
+            new_val = event.data["new_state"].attributes[attribute]
+            if "old_state" in event.data and attribute in event.data["old_state"].attributes:
+                old_val = event.data["old_state"].attributes[attribute]
+            else:
+                old_val = None
+
+            if old_val != new_val:
+                val_change = False
+            else:
+                val_change = True
+
+            new_vars = {f"{var_name}.{attribute}": new_val, f"{var_name}.old.{attribute}": old_val}
+            func_args = {
+                "trigger_type": "state",
+                "var_name": f"{var_name}.{attribute}",
+                "value": new_val,
+                "old_value": old_val,
+                "change": val_change,
+                "context": event.context,
+            }
+            await State.update(new_vars, func_args)
 
     async def hass_started(event):
         _LOGGER.debug("adding state changed listener and starting global contexts")
--- a/custom_components/pyscript/eval.py
+++ b/custom_components/pyscript/eval.py
@@ -455,7 +455,7 @@ class EvalFunc:
         kwarg_check = {
             "task_unique": {"kill_me"},
             "time_active": {"hold_off"},
-            "state_trigger": {"state_hold", "state_check_now"},
+            "state_trigger": {"state_hold", "state_check_now", "only_on_change"},
         }
         for dec_name in trig_args:
             if dec_name not in kwarg_check and "kwargs" in trig_args[dec_name]:
--- a/custom_components/pyscript/state.py
+++ b/custom_components/pyscript/state.py
@@ -77,9 +77,13 @@ class State:
 
         for var_name in var_names if isinstance(var_names, set) else {var_names}:
             parts = var_name.split(".")
-            if len(parts) != 2 and len(parts) != 3:
+            if len(parts) == 2:
+                state_var_name = f"{parts[0]}.{parts[1]}"
+            elif len(parts) == 3:
+                state_var_name = f"{parts[0]}.{parts[1]}.{parts[2]}"
+            else:
                 continue
-            state_var_name = f"{parts[0]}.{parts[1]}"
+
             if state_var_name not in cls.notify:
                 cls.notify[state_var_name] = {}
             cls.notify[state_var_name][queue] = var_names
--- a/custom_components/pyscript/trigger.py
+++ b/custom_components/pyscript/trigger.py
@@ -538,6 +538,7 @@ class TrigInfo:
         self.state_trigger_kwargs = trig_cfg.get("state_trigger", {}).get("kwargs", {})
         self.state_hold_dur = self.state_trigger_kwargs.get("state_hold", None)
         self.state_check_now = self.state_trigger_kwargs.get("state_check_now", False)
+        self.only_on_change = self.state_trigger_kwargs.get("only_on_change", True)
         self.time_trigger = trig_cfg.get("time_trigger", {}).get("args", None)
         self.event_trigger = trig_cfg.get("event_trigger", {}).get("args", None)
         self.state_active = trig_cfg.get("state_active", {}).get("args", None)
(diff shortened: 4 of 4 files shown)


dlashua:
This test code seems to do all the right things:

```python
@time_trigger('startup')
def set_test_state():
    task.unique('set_test_state')
    while True:
        state.set('binary_sensor.test', 'on')
        task.sleep(15)
        state.set('binary_sensor.test', 'off')
        task.sleep(15)

@time_trigger('startup')
def set_test_attrib():
    task.unique('set_test_attrib')
    cnt = 0
    while True:
        cnt = cnt + 1
        if cnt > 10:
            cnt = 1
        state.set('binary_sensor.test', cnt=cnt)
        task.sleep(1)


@state_trigger('binary_sensor.test != "never_happens"')
def report_state_always(**data):
    log.info(data)

@state_trigger('binary_sensor.test.cnt != "never_happens"')
def report_attrib_always(**data):
    log.info(data)

@state_trigger('binary_sensor.test == "on" and binary_sensor.test.old == "off"')
def report_state_on_from_off():
    log.info('Is ON, was OFF')

@state_trigger('binary_sensor.test.cnt == 5 and binary_sensor.test.old.cnt == 4')
def report_cnt_5_from_4():
    log.info('CNT is 5, was 4')

@state_trigger('binary_sensor.test')
def report_state_by_eid(**data):
    log.info(data)

@state_trigger('binary_sensor.test.cnt')
def report_attrib_by_eid(**data):
    log.info(data)
```

dlashua:
@craigbarratt I don't know why these two tests are failing. I don't even know that it's related or how to tell what went wrong. Can you shed some light. (my knowledge of python tests is limited)

dlashua:
An issue of sorts.

This code...
```python
@state_trigger('binary_sensor.test')
def report_state_by_eid(**data):
    log.info(data)
```

... before this PR meant to trigger on ANY change (state OR attribute) to `binary_sensor.test`. Now, with this PR, without adding `only_on_change=False` it only triggers if the actual STATE changes. This change will probably break some existing scripts.

What would be best is for a comparison style trigger (`binary_sensor.test == "on"`) to have `only_on_change=True` but a non-comparison (just `binary_sensor.test`) to have `only_on_change=False` by default. But that gets weird.

dlashua:
Another option, which will create another possible collision in the `binary_sensor.test` namespace.

`@state_trigger('binary_sensor.test')` means trigger on any change to state OR any attribute. `func_args` will still just show `binary_sensor.test` and it's state for `value` and `old_value`. 

`@state_trigger('binary_sensor.test.some_attrib')` means trigger on a change to that one attribute. `func_args` will show the attribute and the old and new values of the attribute.

`@state_trigger('binary_sensor.test.state')` means only trigger on a change to STATE. `func_args` will show `binary_sensor.test` as it does now.

So, we trigger on ANY change unless the user specifically indicates (via `.state`) that they only want to check the state.

Or... we add another kwarg of "state_only" or something to mean just the state.

craigbarratt:
For some reason I thought that the `state_changed` event didn't have the new and old state attributes, but indeed it does.  That's really good news.

I agree that `@state_trigger` shouldn't evaluate the expression if some unmentioned attribute changes.

I'd recommend we do the following.  The `@state_trigger` expression is only evaluated if variables that are mentioned in the expression (whether its value `domain.entity`,  attribute like `domain.entity.attr`, or virtual ones like `domain.entity.old`) have changed. That seems the most natural approach, and solves the current issue.  Since we keep track of which state variables or attributes are mention in the trigger expression, the notify code in `state.py` could do the comparisons to see what has changed.

I'd prefer the any-change trigger `@state_trigger("domain.entity")` to mean only any change to the value, not attributes.  While that would be a breaking change, I doubt many people are depending on that triggering on attribute changes right now. But I could be convinced otherwise - perhaps it could be a kwargs flag that defaults to the old behavior as you've done in #76. We could also support `@state_trigger("domain.entity.attr")` to mean any change to that attribute.

That's leaves a couple of open questions:
* how to specify an any-change trigger.  It could be a kwarg option as just mentioned, some virtual attribute or a different decorator.  I haven't settled on the preferred method here - maybe the kwarg approach is best.
* we now have an issue that a trigger could involve multiple variables changes if there are multiple attributes that change and the trigger expression mentions them.  What should we set the function arguments `var_name`, `value` and `old_value` to in that case?
* do we support `old` attributes in the trigger expression, of the form `domain.entity.old.attr`?  I'd say yes.

craigbarratt:
One way to generalize and address the 2nd bullet (what to pass as `value` to the trigger function) is to make state variables evaluate to class instances that do the right thing, rather than fake them as being strings, but attributes still work.

This addresses a comment you made in #30: 
```python
    log.info(hasattr(input_boolean.test_1,'editable'))  # False
    log.info(input_boolean.test_1.editable)  #True
```
If we support that, it would be backwards compatible (the instance on its own would magically evaluate to a string), but would actually be an instance that provides attributes etc.

Then `value` and `old_value` could be the trigger state variables, and then inside the trigger function things like` value.attr` or `old_value.attr` would do the right thing, and would be based on the actual trigger state change, and not be subject to a race condition if the current attributes are fetched (which might have changed since the trigger), which is what happens now.

Then the `@state_trigger` function would be passed `domain.entity` and `domain.entity.old` set the real state objects, rather than just the strings currently, with attribute accesses falling back to fetching, which is prone to race conditions.

dlashua:
As this PR is now, it works as you describe, minus the "state variables are class instances" bit. A state trigger for binary_sensor.test will only trigger if the value changes. A state trigger for binary_sensor.test.cnt will only trigger if that attribute changes. And the func_args have var_name set to the state variables or the attribute variable based on the trigger, with value and old_value being set as expected. 

Making "state variables class instances" gets into the tricky parts of pyscript code I don't fully understand yet (the Ast bits). So I'm not quite qualified to do that if that's the direction you want to take.

craigbarratt:
I'm also not yet sure how to make "state variables class instances".  It's quite tricky and I need to think about it some more.  If it's possible I think that's the way to go, although it's secondary to the bigger issue of only checking the state trigger when a referenced value changes.  It does make the `value` and `old_value` arguments work in a more elegant way, and would simplify the implementation.

The PR definitely got me thinking about how to do this.  Currently, however, it potentially calls `State.update()` multiple times - once if the value changes, and once for each attribute that has changed, independent of whether those are used in the `@state_trigger`.  It actually needs to call it just once, so that the trigger function is only evaluated once.  The logic about whether a referenced state variable or attribute has changed (ie, the ones mentioned in the `@state_trigger`) needs to move to the notify logic in `state.py`.

Let me think about the "state variables class instances" some more.

dlashua:
Ah ha. My test code was too narrow to see if the trigger function was being called more than once.

I'll see if I can make these changes. Thank you!

dlashua:
This PR essentially makes entity attributes act like entity states. Effectively, it issues its own version of a state_changed event for every state AND attribute. I think having a state change for every attribute is desirable. In fact, I wish Home Assistant did that already on its own. It would certainly elimate the tons of "feature requests" they've gotten for things that don't support attributes in the same way they support states. 

Native Home Assistant Automation TEMPLATE Triggers don't have this issue because they approach the problem differently. If I write a Native Automation with two separate (NON TEMPLATE) triggers -- one for binary_sensor.test and one for binary_sensor.test.cnt -- it WILL trigger twice, by design.

However, if I instead use a single HASS Template Trigger in my automation, it would only fire once. This is because the HASS Automation keeps track of the True/False state of the Template Trigger, and only triggers if it changes from False to True. Future state_changed events that STILL evaluate to True do not cause a subsequent firing of the automation. With this in place, it doesn't matter if the changes come in one state_changed event or two, because the True/False evaluation of the Template will still only change once. 

Personally, I PREFER the way pyscript works. I want it to trigger every time there is a change (not just when the False to True line is crossed). I can decide in my function if I want to do anything about that change. I can even keep track of False/True if I want, and only take action when it changes from False to True. 

Something like a kwarg of "only_once=True" on state_trigger could invoke this False to True trigger logic if you wanted that to be in pyscript. Users can already do this themselves in code, but it wouldn't hurt to do it for them. I think it's common enough in general that it'd be nice to have, but not common enough for me personally that it's any trouble to just do it myself when I need it.

As it stands now, a state_trigger is, essentially, this:

```python
@state_trigger('binary_sensor.test == "on" and int(sensor.cnt) > 5')
```

```yaml
- trigger:
    platform: state
    entity_id:
      - binary_sensor.test
      - sensor.cnt
  condition:
    platform: template
    value_template: "{{ is_state('binary_sensor.test', 'on') and states('sensor.cnt')|int > 5 }}"
```

Though I haven't tested it, I imagine the above Home Assistant Automation would also trigger twice if the changes came close enough together.

For me, pyscript has the perfect blend of functionality and terseness of expression. The things I do most often are the most terse to write, the things I need less often are a bit more verbose but still possible. And having the trigger fire twice isn't a problem since I can account for it in my trigger function.

All of this to say, I think the trigger function running twice in race conditions, while not desirable, is okay. And it's happening now anyway without this PR so, while this PR doesn't fix that problem, it also doesn't make it much worse. If someone NEEDS to have just ONE trigger for any given entity this syntax is available: 

`@state_trigger('binary_sensor.test', only_on_change=False)`

dlashua:
We can pass all of "old_state" and "new_state" from `state_changed` into the func_args. This would give users the ability to see what the old attribute value was in their code by doing:

```
@state_trigger('binary_sensor.test')
def a(**data):
  log.info(data['old_state'].attributes['cnt'])
```

"value" and "old_value" in `func_args` becomes redundant because the data is available in "old_state" and "new_state" as well.

But this would resolve the issue of not knowing what the old attribute value was inside the trigger function. This isn't the issue mentioned in #73, but is still a problem pyscript users have had trouble navigating.

dlashua:
I can make the call to `State.update()` only happen once per state_changed event, but `State.update()` breaks that into several `queue.put()`s anyway, so the end result would still be, potentially, multiple trigger function fires. And with a single call to `State.update()` we lose the ability to customize func_args for the values of the attribute.

If we go back to only notifying on state variables (domain.entity), we'd get back to only one trigger function fire per state_changed event. And users could still still test for binary_sensor.test.old.cnt in the state_trigger because we could pass those variables in to `State.update()`. And a check before the trigger function fired could make sure that SOMETHING changed (state or attribute) but we'd still see trigger function fires when an attribute changed when the user only cares about the state.

If we want to limit trigger function fires to only when the thing we care about actually changes value, then the most straight forward solution seems to be to notify on each attribute.

dlashua:
Perhaps we can call `State.update()` only once. We include all of old_state and new_state in the notification. Then trigger.py can compare them all, determine what changed, look at the variables in state_trigger and build func_args on its own based on what the user cares about.  However, if I have @state_trigger('int(sensor.test) > 5 and sensor.test.cnt > 5 and sensor.test.other > 5') and all three values change in the same state_changed event, which one would we trigger on? 

Perhaps we always trigger with var_name = 'sensor.test', even if it was an attribute that changed?

We could include a dict like this in func_vars:

```yaml
changes:
  sensor.test:
    old: 4
    new: 5
  sensor.test.cnt:
    old: 3
    new: 7
  sensor.test.other:
    old: 4
    new: 15
```

And we only trigger the function if they mentioned that specific thing in state_trigger. So a change to sensor.test.cnt with state_trigger('sensor.test') would not call the function.

dlashua:
replaced by #82

PR #82, opened 2020-11-05
state_trigger only triggers on value changes (using StateVar)

dlashua:
This works for me. Needs more testing.

Does not implement domain.entity.* functionality. Suggesting we save that for a future PR.


dlashua:
OMG, I fixed a test AND caught a bug because of a test! GO TESTS!

dlashua:
entity.domain.* is not valid python, so it's a bunch of extra hurdles to jump through to make this the way to designate "all attributes".

I suggest we either use something else -- entity.domain.all, perhaps -- or we indicate this functionality with a kwarg or we just don't implement it at all and require users to be specific about what they want to trigger on. Thoughts?

craigbarratt:
This is look great.

The trigger on any change values (eg: `"domain.entity"` or `"domain.entity.attr"`) aren't passed through the Python interpreter - they are separated and matched independent of the remaining expression.  So that gives us flexibility in what syntax to use.  I think it should just be a couple of lines of code to handle the `.*` case.  When the code matches for plain entity or entity attribute (assuming the regexp allows the optional 3rd part to also match `*`), then if the 3rd part is `"*"` we simply check if any attribute has changed.

The drawback with `.all` is it is another case of potentially obscuring a real attribute; you wouldn't be able to trigger off any changes to an attribute called `domain.entity.all`.  So I'm still in favor of using `"domain.entity.*"` to denote any attribute change.

dlashua:
Excellent. I can see if I can figure out where to put that, but the deeper it gets into the Ast bits, the harder it is for me to decipher. Perhaps if you get some time you can figure out where to put it? Right now, if I put "domain.entity.*" into a state_trigger, I get an error like this:

```
2020-11-05 19:36:32 ERROR (MainThread) [custom_components.pyscript.file.bigtest.report] Exception in <file.bigtest.report @state_trigger()> line 1:
    binary_sensor.test.*
                       ^
SyntaxError: unexpected EOF while parsing (file.bigtest.report @state_trigger(), line 1)
```

craigbarratt:
You need to update `STATE_RE` in `trigger.py` to match all three cases.  Then that will cause all the "any change" triggers to get removed from the expression and saved in `state_trig_ident_any`.  Then the trigger code needs to check `state_trig_ident_any` to see if any of the "any change" variables/attributes there have actually changed or not.

dlashua:
okay. I think I can manage that.  I'll give it a shot now.

dlashua:
@craigbarratt limited testing shows domain.entity.* feature to be working.

custom_components/pyscript/trigger.py
```diff
@@ -812,6 +822,41 @@ async def trigger_watch(self):
...
                     task_unique_func = Function.task_unique_factory(action_ast_ctx)
 
+                #
+                # check for changes to state or attributes
+                #
+
+                # if "value" not in func_args, then we are state_check_now
+                if "value" in func_args and notify_type == 'state':
+                    trig_ident_change = False
+
```
craigbarratt, line 832: Might be easier to extract `var_name`, `value`, `old_value` as variables, just to reduce the repetitive code below.

custom_components/pyscript/trigger.py
```diff
@@ -739,7 +739,17 @@ async def trigger_watch(self):
...
+                    any_match = False
+                    if "var_name" in func_args:
+                        for check_var in self.state_trig_ident_any:
+                            if check_var == func_args['var_name']:
+                                any_match = True
+                                break
+                            if check_var.startswith(f"{func_args['var_name']}."):
+                                any_match = True
+                                break
+                    if not any_match:
```
craigbarratt, line 752: Somehow my longest and most important comment got lost, so repeating it here.  Basically all the checks below should move up above this line, so that the trigger function is only evaluated if necessary.  Specifically, it should be skipped altogether if any of the three "any" cases are true:
- state value change when `state_trig_ident_any` contains `var_name`
- attribute change for any entry in `state_trig_ident_any` of the form `var_name.attr`
- any attribute change if an entry in `state_trig_ident_any` is of the form `var_name.*`

Next, you then should move the code that checks `self.state_trig_ident` to see if any of mentioned variables have changed.  Only if that is the case should the trigger function be evaluated.

custom_components/pyscript/trigger.py
```diff
@@ -812,6 +822,41 @@ async def trigger_watch(self):
...
+
+                # if "value" not in func_args, then we are state_check_now
+                if "value" in func_args and notify_type == 'state':
+                    trig_ident_change = False
+
+                    # determine if the catchall has been requested in state_trig_ident_any
+                    catch_all_entity = f"{func_args['var_name']}.*"
+                    if catch_all_entity in self.state_trig_ident_any:
+                        # catch all has been requested, check all attributes for change
+                        for attribute in func_args['value'].__dict__:
```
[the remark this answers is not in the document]
craigbarratt: However, that one-liner is maybe too aggressive - it only works if all the attribute values are immutable (ie, hashable).  So I don't recommend we use that.

You could loop over these keys:
```
(set(value.__dict__.keys()) | set(value.__dict__.keys())) - {"last_updated", "last_changed"}
```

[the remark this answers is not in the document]
dlashua: Yes! That should be much better!

dlashua:
@craigbarratt I think I covered everything. Let me know if I missed something.

custom_components/pyscript/trigger.py
```diff
@@ -644,6 +644,112 @@ def start(self):
             self.task = Function.create_task(self.trigger_watch())
             _LOGGER.debug("trigger %s is active", self.name)
 
+    def ident_any_values_changed(self, func_args):
+        """Check for changes to state or attributes on ident any vars"""
+        value = func_args.get('value')
+        old_value = func_args.get('old_value')
```
craigbarratt, line 650: Should the prior two lines be below the next `if` statement?  I'm not sure if `value` is set if there is no `var_name`, or add `None` defaults to each `get`.

custom_components/pyscript/trigger.py
```diff
@@ -644,6 +644,112 @@ def start(self):
...
+                var_pieces = check_var.split('.')
+                if len(var_pieces) == 3 and f"{var_pieces[0]}.{var_pieces[1]}" == var_name:
+                    if var_pieces[2] == "*":
+                        # catch all has been requested, check all attributes for change
+                        all_attributes = set()
+                        if value is not None:
+                            all_attributes |= set(value.__dict__.keys())
+                        if old_value is not None:
+                            all_attributes |= set(old_value.__dict__.keys())
+                        all_attributes -= {"last_updated", "last_changed"}
```
craigbarratt, line 678: The RHS should be `STATE_VIRTUAL_ATTRS` from `state.py`.

custom_components/pyscript/trigger.py
```diff
@@ -644,6 +644,112 @@ def start(self):
...
+        _LOGGER.debug(
+            "%s no ident_any change detected",
+            self.name,
+        )
+        return False
+
+    def ident_values_changed(self, func_args):
+        """Check for changes to state or attributes on ident vars"""
+        value = func_args.get('value')
+        old_value = func_args.get('old_value')
```
craigbarratt, line 709: Same comment as above

custom_components/pyscript/trigger.py
```diff
@@ -644,6 +644,112 @@ def start(self):
...
+            return False
+
+        for check_var in self.state_trig_ident:
+            if check_var in self.state_trig_ident_any:
+                _LOGGER.debug(
+                    "%s ident change skipping %s because also ident_any",
+                    self.name,
+                    check_var,
+                )
+                continue
```
craigbarratt, line 726: Is this whole `if` clause necessary?  In this case `ident_any_values_changed()` would already have returned `True`.

custom_components/pyscript/trigger.py
```diff
@@ -22,7 +22,7 @@
 _LOGGER = logging.getLogger(LOGGER_PATH + ".trigger")
 
 
-STATE_RE = re.compile(r"[a-zA-Z]\w*\.[a-zA-Z]\w*$")
+STATE_RE = re.compile(r"[a-zA-Z]\w*\.[a-zA-Z]\w*(\.[a-zA-Z\*]?[a-zA-Z]*)?$")
```
craigbarratt, line 25: This RE isn't quite right, since it allows `x.y.*attr` and doesn't allow underscores or digits in the attribute name.

Here's a corrected version:
```
STATE_RE = re.compile(r"[a-zA-Z]\w*\.[a-zA-Z]\w*(\.(([a-zA-Z]\w*)|\*))?$")
```

craigbarratt:
This is looking great - all the logic looks correct to me.  I just added a few cosmetic comments.

I'm happy to merge after your next changes, and then I can add the same features to `task.wait_until()`.  Or you are welcome to do that too.  I think you'll need to make the two functions global so they can be called from both places.

custom_components/pyscript/trigger.py
dlashua: `get` defaults to None if a default is not specified.

custom_components/pyscript/trigger.py
dlashua: My thought process was that ident_any_values_changed() checked it, but the value did not change. Therefore, it didn't return True. So, in ident_values_changed() we can skip it (and save some CPU) since they were already checked. If we take the if out, since it didn't return True before, it won't return True now either.

dlashua:
As yes, I can do the wait_until part as well. These functions will need to not only be global, but also take self.state_trig_ident_any, etc, as an additional parameter since it won't be there as "self".

dlashua:
@craigbarratt limited tested shows task.wait_until() working as expected. 

I don't use that feature much, so I wasn't sure what to test. I tested waiting for a state to change (and ensuring the wait didn't end because an attribute changed), wait for an attribute to change, waiting for "*", and waiting for an attribute to be a certain thing. They all seemed to work correctly.

craigbarratt:
This looks good to me.  I'm happy to merge if you don't have any other changes.

I'll add some tests after I merge.

dlashua:
I think it's ready to go. Thanks!

PR #83, opened 2020-11-05
don't allow additional kwargs to task.wait_until (fixes #80)

### LINKED ISSUE #80: task.wait_until() shouldn't allow unrecognized keyword arguments
The task.wait_until() function ought to cause an error if an unrecognized keyword argument is specified, but it currently ignores such argument.  That makes it hard to catch user errors such as misspelling "state_check_now" as "state_check".

Pyscript Version: 0.32
HA version: 0.117.2

To reproduce:
   Run a script containing:
  ` trig_info = task.wait_until(bozo=5)`

Expected behavior:
  Abort the script and show an exception and traceback in the home assistant log

Observed behavior:
  Acts like:
  `trig_info = task.wait_until()`

I suspect the cause is that wait_until() in pyscript's trigger.py declares an argument **kwargs (which it doesn't use).

dlashua:


dlashua:
Fixed in master.

PR #85, opened 2020-11-06
Add support for service blocking

raman325:
I have created several functions in `pyscript` that make calls to other services. I could use `task.wait_until` to block execution until the service is finished running, but it seems more convenient to be able to specify that in the service call directly. As such, I have added support for the `blocking` and `limit` parameters which will give users more control over the service call behavior.

I am unsure how to test this - I'd be happy to add tests if you have some ideas on the best way to do that.

### DIFF
--- a/custom_components/pyscript/function.py
+++ b/custom_components/pyscript/function.py
@@ -190,13 +190,18 @@ class Function:
     async def service_call(cls, domain, name, **kwargs):
         """Implement service.call()."""
         curr_task = asyncio.current_task()
-        if "context" in kwargs and isinstance(kwargs["context"], Context):
-            context = kwargs["context"]
-            del kwargs["context"]
-        else:
-            context = cls.task2context.get(curr_task, None)
+        hass_args = {}
+        for keyword, typ, default in [
+            ("context", Context, cls.task2context.get(curr_task, None)),
+            ("blocking", bool, None),
+            ("limit", float, None),
+        ]:
+            if keyword in kwargs and isinstance(kwargs[keyword], typ):
+                hass_args[keyword] = kwargs.pop(keyword)
+            elif default:
+                hass_args[keyword] = default
 
-        await cls.hass.services.async_call(domain, name, kwargs, context=context)
+        await cls.hass.services.async_call(domain, name, kwargs, **hass_args)
 
     @classmethod
     async def service_completions(cls, root):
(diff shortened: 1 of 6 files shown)


custom_components/pyscript/function.py
```diff
@@ -190,13 +190,18 @@ def service_has_service(cls, domain, name):
...
+        for keyword, typ, default in [
+            ("context", Context, cls.task2context.get(curr_task, None)),
+            ("blocking", bool, None),
+            ("limit", float, None),
+        ]:
+            if keyword in kwargs and isinstance(kwargs[keyword], typ):
+                hass_args[keyword] = kwargs.pop(keyword)
+            elif default:
+                hass_args[keyword] = default
+
```
craigbarratt, line 203: This is a nice compact way to process the `kwargs`.  I was thinking we should log an error if the keyword argument doesn't match the type, but since these potentially override service parameters with same name, we should probably just silently pass them along as you are doing.  So this look as is.

raman325:
I figured out how to test at least the changes in `pyscript.function`. Please hold off on merging while I add tests.

craigbarratt:
This looks good - I don't have any suggested changes.

For testing, it would be great if you could add some.  These are quite tricky though because you can't actually wait any noticeable amount of time, and there are race conditions.  Here's what I'd recommend.

For `blocking`, the test could use these steps:
* copy the initialization where we handshake the `hass started` event to make sure everything is up and running
* in the pyscript startup function:
    - set `pyscript.var1 = 1`
    - make a service call to another pyscript function with `blocking=True`
    - that second pyscript function sets `pyscript.var1 = 2`
    - immediately after the service call returns, send `pyscript.var1` back through the `pyscript.done` queue
    - the test code asserts that `pyscript.var1 = 2`
    - repeat the test for other type of service call (eg: both `service.call()` and `pyscript.func()`)

I don't see a good way to test `blocking=False` since there are race conditions.

For `limit`, do a test similar to the above, but the second function just does a long task.delay() and the caller uses `limit = 1e-6` and `blocking=True`.  Just the fact the service call returns means limit is working.

raman325:
Let me know what you think of the tests I added. We don't actually have to test that the function blocks since that is Home Assistant's job, we just need to make sure that the appropriate params get passed to `async_call`

craigbarratt:
Do you think this is ready now?  I'm happy to try adding the couple of tests I proposed.  Your tests look good too.

raman325:
Yes I feel good about this PR now, but if you would like to add some tests that actually test the blocking please do. I do get a little lost with what's sync and async in pyscript so I tend to make my tests very targeted.

craigbarratt:
Great - thanks.  Merged.

craigbarratt:
I added some additional tests in a25063d.

raman325:
Great, thank you!

PR #88, opened 2020-11-10
await in deprecated get_attr

dlashua:
This PR will prevent the error:

`TypeError: argument of type 'coroutine' is not iterable`

In the following code:

```
@time_trigger('startup')
def test():
    a = state.get_attr('sun.sun')

    for attrib in a:
        log.info(attrib)
```

### DIFF
--- a/custom_components/pyscript/state.py
+++ b/custom_components/pyscript/state.py
@@ -326,7 +326,7 @@ class State:
     async def get_attr(cls, var_name):
         """Return a dict of attributes for a state variable - deprecated."""
         _LOGGER.warning("state.get_attr() is deprecated: use state.getattr() instead")
-        return cls.getattr(var_name)
+        return await cls.getattr(var_name)
 
     @classmethod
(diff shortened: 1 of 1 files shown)


craigbarratt:
Oh no!  I committed the sin of making a "simple" change without adding a test, and right before a release.  Sigh.

craigbarratt:
At least it will get people to use the new function...

dlashua:
Haha. I wouldn't sweat it. It was time for me to start using getattr anyway. :) I had some "old" (relatively speaking, anyway) code that I hadn't worked on in a while. So I was surprised to see it start throwing errors. :)

dadler:
FYI, this temporary breaking change just bit me too (in the 1.0.0 release).  If the fix isn't released soon, I wonder if it'd be worth temporarily mentioning the bug in the reference documentation and release notes in HACS, which currently indicate get_attr is deprecated but works?

PR #91, opened 2020-11-17
Install packaging module if not found

### LINKED ISSUE #90: Component error: pyscript - No module named 'packaging'
Hi
Can you help me fixind this issue? 
I have upgraded via HACS from 0,xx to 1.0.0 but now I am stuck with this log.

Running HA in virtual env python_version 3.7.3

Thanks

raman325:
Resolves #90

### DIFF
--- a/custom_components/pyscript/requirements.py
+++ b/custom_components/pyscript/requirements.py
@@ -4,7 +4,6 @@ import logging
 import os
 import sys
 
-from packaging.version import Version
 from pkg_resources import Requirement
 
 from homeassistant.loader import bind_hass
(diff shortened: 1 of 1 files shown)


craigbarratt:
Thanks for quickly looking into and addressing #90.

One question - is it possible to fix via just adding packaging to `requirements` in `custom_components/pyscript/manifest.json`?

raman325:
yes, that may be a better approach, and I don't know why I didn't think of that

raman325:
The only rub there will be that for certain installations of Home Assistant (I know the Docker container already has packaging), packaging is already installed, and I'm not clear on whether or not pinning it to a version will break something in the future

craigbarratt:
Good point - it looks like `requirements` in `manifest.json` is always pinned, which is problematic for the reason you mention.  I haven't looked at the code to see whether pinning is required, but it seems risky even if pinning wasn't required since all components I checked do pin.

It does seem like this PR is the best fix.

craigbarratt:
Thanks!

PR #100, opened 2020-11-30
use list for unique_task2name

### LINKED ISSUE #99: cls.unique_task2name should hold lists
example code:

```python
def whatever():
  task.unique('a')
  task.unique('b')
```

at line 170 of function.py, cls.unique_task2name[curr_task] will be equal to "b" after this is complete.

Therefore, when cleanup happens at line 301, only cls.unique_name2task['b'] is cleaned. 'a' still shows this task holding.

Any unique name can only hold one task. But a task should be able to hold multiple unique slots.

Alternately, the kill_me logic at line 151 could check to see if the task in the slot is "finished" and ignore it if it is. 

I use multiple task.unique statements in the same function because I have two functions that run against one another with delays... an on/off scenario. If the thing is ON, then it uses task.unique to kill any pending off task, and then uses it again to set its own pending on task. The off function similarly kills any pending on task before setting its own pending off task.

dlashua:
Fixes #99

### DIFF
--- a/custom_components/pyscript/function.py
+++ b/custom_components/pyscript/function.py
@@ -166,13 +166,11 @@ class Function:
                 if name in cls.unique_name2task:
                     task = cls.unique_name2task[name]
                     if task in cls.unique_task2name:
-                        if name in cls.unique_task2name[task]:
-                            cls.unique_task2name[task].remove(name)
+                        cls.unique_task2name[task].discard(name)
                 cls.unique_name2task[name] = curr_task
                 if curr_task not in cls.unique_task2name:
-                    cls.unique_task2name[curr_task] = []
-                if name not in cls.unique_task2name[curr_task]:
-                    cls.unique_task2name[curr_task].append(name)
+                    cls.unique_task2name[curr_task] = set()
+                cls.unique_task2name[curr_task].add(name)
 
         return task_unique
 
--- a/custom_components/pyscript/function.py
+++ b/custom_components/pyscript/function.py
@@ -163,6 +163,11 @@ class Function:
                     except asyncio.CancelledError:
                         pass
             if curr_task in cls.our_tasks:
+                if name in cls.unique_name2task:
+                    task = cls.unique_name2task[name]
+                    if task in cls.unique_task2name:
+                        if name in cls.unique_task2name[task]:
+                            cls.unique_task2name[task].remove(name)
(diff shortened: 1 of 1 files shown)


custom_components/pyscript/function.py
```diff
@@ -163,8 +163,16 @@ async def task_unique(name, kill_me=False):
                     except asyncio.CancelledError:
                         pass
             if curr_task in cls.our_tasks:
+                if name in cls.unique_name2task:
+                    task = cls.unique_name2task[name]
+                    if task in cls.unique_task2name:
+                        if name in cls.unique_task2name[task]:
+                            cls.unique_task2name[task].remove(name)
```
craigbarratt, line 170: It would be more efficient to use a set, in which case the last two lines can just be a single-line `discard` (which is a no-op if not present)

custom_components/pyscript/function.py
```diff
@@ -163,8 +163,16 @@ async def task_unique(name, kill_me=False):
...
                 cls.unique_name2task[name] = curr_task
-                cls.unique_task2name[curr_task] = name
+                if curr_task not in cls.unique_task2name:
+                    cls.unique_task2name[curr_task] = []
```
craigbarratt, line 173: This would be `set()` instead of `[]`

craigbarratt (commented):
This looks great.  My only suggestion is to use a set instead of list, which is more efficient if we had a lot of names, and also `add` and `discard` don't require testing beforehand.

dlashua:
I considered a set because it makes the values unique automatically. But, having never used one before, when I looked up the differences between a list and set I read that items in a set could not be changed or replaced. So I went with a list because I knew it would work (i.e. I chickened out on having to debug it if it didn't work as I expected).

I'll convert to a set.

craigbarratt:
`set` elements have the same restriction as `dict` keys - they have to be immutable (so their hashed value cannot change).  So `set` elements and `dict` keys can't be things like lists, sets, dicts, or objects, since they can be changed.

A `set` is just like a `dict` where there isn't a value and the set members are the dict keys.

craigbarratt:
Looks great; fixes #99.

dlashua:
> A `set` is just like a `dict` where there isn't a value and the set members are the dict keys.

well... TIL about object hashes in python. I've never tried to use a dict as a dict key before, but now I know why I can't. Thanks!

PR #105, opened 2020-12-01
add @mqtt_trigger

### LINKED ISSUE #98: Feature Request: MQTT trigger
**Note: the original tile of the question was "[question] how to run a always-on script?"**

I am converting my automation from AppDaemon to `pyscript` and some of them depend on MQTT (as a client). A typical case is a switch emitting to MQTT topic I want to listen to and act accordingly.

This means that such a script is not triggered by anything but runs in the background, responds to some events it manages itself (MQTT messages in my case), and then contacts HA for some actions (switching on a light for instance).

**Is it possible to run such a script?** The only cases I saw in the documentation are ones that react to something (either time based, or coming from HA)

One solution would be to use HA to manage the MQTT messages, but there are problems with this in some cases (see https://community.home-assistant.io/t/how-to-get-the-name-of-the-actual-state-topic-in-an-mqtt-sensor/248541 for an example). Using HA is doable but I would prefer to avoid automations there and move everything to pyscript for consistency.

dlashua:
This is a quick first attempt. The simplest of test scripts seems to work:

```python
@time_trigger('startup')
def mqtt_counter():
    task.unique('mqtt_counter')
    cnt = 0
    while True:
        cnt = cnt + 1
        log.info(cnt)
        mqtt.publish(topic='pyscript/test', payload=cnt)
        mqtt.publish(topic='pyscript/testx10', payload=(cnt * 10))
        task.sleep(1)

@mqtt_trigger('pyscript/test')
def mqtt_recv(**data):
    log.info(f'got one {data}')

@mqtt_trigger('pyscript/test', 'topic == "notthis"')
def mqtt_never(**data):
    log.info(f'this should never happen {data}')

@mqtt_trigger('pyscript/test', 'int(payload) % 2 == 0')
def mqtt_even(**data):
    log.info(f'EVEN {data}')

@mqtt_trigger('pyscript/#')
def mqtt_wildcard(**data):
    log.info(f'got one from wildcard {data}')

@mqtt_trigger('pyscript/test', 'payload == "5"')
def mqtt_5_then_10():
    log.info('got 5, waiting for 10')
    task.wait_until(mqtt_trigger=['pyscript/test', 'payload == "10"'])
    log.info('got 10')
```

Tests are needed. Though... I'm not sure how to write tests without an mqtt server to test against?

Docs are needed.

Feel free to make edits directly in my repo even if you are unable to test them.

Will fix #98 .

### DIFF
--- a/custom_components/pyscript/__init__.py
+++ b/custom_components/pyscript/__init__.py
@@ -32,6 +32,7 @@ from .const import (
 )
 from .eval import AstEval
 from .event import Event
+from .mqtt import Mqtt
 from .function import Function
 from .global_ctx import GlobalContext, GlobalContextMgr
 from .jupyter_kernel import Kernel
@@ -118,6 +119,7 @@ async def async_setup_entry(hass, config_entry):
 
     Function.init(hass)
     Event.init(hass)
+    Mqtt.init(hass)
     TrigTime.init(hass)
     State.init(hass)
     State.register_functions()
--- a/custom_components/pyscript/eval.py
+++ b/custom_components/pyscript/eval.py
@@ -278,11 +278,13 @@ class EvalFunc:
             "time_trigger",
             "state_trigger",
             "event_trigger",
+            "mqtt_trigger",
         }
         trig_decorators = {
             "time_trigger",
             "state_trigger",
             "event_trigger",
+            "mqtt_trigger",
             "state_active",
             "time_active",
             "task_unique",
@@ -393,6 +395,7 @@ class EvalFunc:
         #
         arg_check = {
             "event_trigger": {"arg_cnt": {1, 2}},
+            "mqtt_trigger": {"arg_cnt": {1, 2}},
             "state_active": {"arg_cnt": {1}},
             "state_trigger": {"arg_cnt": {"*"}, "type": {list, set}},
             "task_unique": {"arg_cnt": {1}},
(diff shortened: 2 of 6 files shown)


craigbarratt:
I reviewed this and it looks great!  I don't have any comments.  Really nice work.

The next two parts are docs and tests.  I realize the tests are not easy to understand, so I can take a crack at that if you want.

dlashua:
@craigbarratt absolutely! I can handle docs if you can work out whatever the tests for this might look like.

dlashua:
Docs have been added.

As it is now, payload_json is only set in func_args if the payload is valid JSON. Otherwise, it doesn't exist. Could that cause problems? Should we set it to None instead when it is unparsable?

craigbarratt:
Looks great.

I think the `payload_json` name is confusing.  To me it sounds like that's the string json form, not the deserialized object. How about `payload_obj`?  (Names like `payload_from_json` or `payload_deserialized` are too long for my liking.)

I'm ok with your current approach of it being undefined if `payload` isn't a valid json string.

dlashua:
Yes. `payload_obj` is closer. It's a weird thing. For instance, let's say the payload is "2". 

A payload will always be a string. So `payload = "2"`. However, `payload_json = 2`... the integer. But, in most cases, it will be an object, so, that's better than what we have. I'll make that change until we can come up with something better.

PR #119, opened 2020-12-14
report errors in time_active

### LINKED ISSUE #118: No error message in log when using illegal time_active statement
Hi,

having an error in an time_active statement does not generate an error. For example this statement does not generate any issue in the log file and prevent the rule from beeing fired:

`@time_active("range(sunrise, sunset")`

Hint: there is one closing bracket missing after the sunset. 

Is it possible to log such errors as it is really painful to find them when they are silently ignored.

Beside this, Kudos for developing this great tool. Thanks and keep going! :-)

dlashua:
Fixes #118


craigbarratt:
Nice - thanks!

Issue #123, opened 2020-12-21
Feature request: Enforce per-app configuration and make global config explicitly global


exponentactivity:
EDIT: My original feature request was already done, see the next comments for discussion of the updated title.

I made a script that calls an API on my local network, and it needs a url and an api key to do that.

The problem is i would like to backup this script to git (with the rest of my config) and be able to share it with others in a usable format, but my url and api is hard coded in clear text.

I looked at [Accessing YAML configuration](https://hacs-pyscript.readthedocs.io/en/stable/reference.html?highlight=access%20hass%20global%20variable#accessing-yaml-configuration) and it seems like an okay solution, but it doesn't seem like it's supposed to contain any sensitive data, and putting it in .gitignore would mean the config is not backed up.

If i could read the api key from secrets.yaml all sensitive data would be kept in a single file.
I think it would be preferable to whitelist the secrets each pyscript script/app has access to in pyscript.config so a malicious app can't access everything. 

Whitelisting could also double as place to map the (arbitrarily named) secrets to the variable inside the script.

`pyscript.config` could look something like this:
```yaml
pyscript:
  allowed_secrets:
    - my_app:
      - my_whitelisted_secret: my_variable_the_secret_is_mapped_to
  apps:
    ...
```

dlashua:
This is already doable in almost exactly the way you specified:

secrets.yaml
```
some_secret: welcome
```

configuration.yaml
```
pyscript:
  secrets:
    some_secret: !secret some_secret
```

in a pyscript
```
@time_trigger('startup')
def startup():
  log.info(pyscript.config['secrets']['some_secret']
```

dlashua:
Keep in mind, in Home Assistant, secrets.yaml doesn't auto reload. So, if you add a secret, you'll need to restart Home Assistant before it's available.

exponentactivity:
That totally solved my problem! Thanks!

I see the documentation for this feature now, i just didn't use the correct search terms. I imagine my use case isn't that out of the ordinary. If you agree i will look into expanding the documentation and send a pull-request? Maybe just some some examples that would make it pop up on search terms related to passwords and sensitive info.

My itch is scratched, but i do think enforcing the recommended use of per-app configuration, and having separate and explicit global configuration would be a great idea, especially if sharing apps becomes widespread, to avoid possible security issues.

Maybe something like:
```yaml
pyscript:
  apps:
    my_app:
      thing_a: this
      thing_b: that
      some_secret: !secret some_secret
[...]

PR #124, opened 2020-12-21
Added an example of using secrets in the config

exponentactivity:
I couldn't find any mention of using vars from secrets.yaml in the docs so i added one in case some one else finds themselves in the same situation as me.

Related to #123


craigbarratt:
Thanks for the PR!

PR #176, opened 2021-03-01
update services.yaml schema to support 2021.3

raman325:
Updated `services.yaml` but wasn't sure if I got the required value right for all parameters of `jupyter_kernel_start`. Can you review and let me know which, if any, need to be updated?

These changes are not required but they do make calling the services from within the UI more friendly.

### DIFF
--- a/custom_components/pyscript/services.yaml
+++ b/custom_components/pyscript/services.yaml
@@ -6,7 +6,7 @@ reload:
   fields:
     global_ctx:
       name: Global Context
-      description: only reload this specific global context (file or app)
+      description: Only reload this specific global context (file or app)
       example: file.example
       required: false
       selector:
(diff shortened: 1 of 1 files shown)


craigbarratt:
`hassfest` failed with this error:
```
Invalid services.yaml: required key not provided @ data['jupyter_kernel_start']['fields']['control_port']['description']
```

raman325:
Thanks for the heads up will push an update shortly. Rereading my PR description I realize I wasn’t clear in what I was asking: fields have a new ‘required’ property that defaults to false. If a field is marked as required=true, the front end won’t let the service be called in dev tools or let the automation be saved in the automation UI if a required field is not filled out. Did I get the required fields right?

craigbarratt:
They look ok to me, but I haven't had time to read more about the service changes.  `jupyter_kernel_start` is not intended for end-users, but the extra detail looks ok.  Currently the kernel generates port numbers starting at 50321, so your minimum number should be fine.  But it's probably better to just pick a lower number that is still unprivileged, like 10240?

PR #332, opened 2022-03-04
fix for state.persist

### LINKED ISSUE #331: state.persist is broken in HA 2022.3
https://github.com/home-assistant/core/pull/66068 changed the signature of `async_restore_entity_added` from taking an entity `str` to taking a `RestoreEntity`

dlashua:
fixes #331

### DIFF
--- a/custom_components/pyscript/entity.py
+++ b/custom_components/pyscript/entity.py
@@ -1,0 +1,9 @@
+from homeassistant.helpers.restore_state import RestoreEntity
+
+
+class PyscriptEntity(RestoreEntity):
+    def set_state(self, state):
+        self._attr_state = state
+
+    def set_attributes(self, attributes):
+        self._attr_extra_state_attributes = attributes
--- a/custom_components/pyscript/entity.py
+++ b/custom_components/pyscript/entity.py
@@ -1,9 +1,14 @@
+"""Entity Classes"""
 from homeassistant.helpers.restore_state import RestoreEntity
 
 
 class PyscriptEntity(RestoreEntity):
+    """Generic Pyscript Entity"""
+
     def set_state(self, state):
+        """Set the state"""
         self._attr_state = state
 
     def set_attributes(self, attributes):
+        """Set Attributes"""
         self._attr_extra_state_attributes = attributes
--- a/custom_components/pyscript/entity.py
+++ b/custom_components/pyscript/entity.py
@@ -1,6 +1,5 @@
 """Entity Classes"""
 from homeassistant.helpers.restore_state import RestoreEntity
-from collections.abc import MutableMapping
 from typing import Any
 from homeassistant.helpers.typing import StateType
 from homeassistant.const import STATE_UNKNOWN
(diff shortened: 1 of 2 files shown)


dlashua:
The failing tests are due to the use of internal class properties externally.

The correct way to handle is it to write a class that extends RestoreEntity and provides methods to set these properties. However, as is, it should resolve the 2022.3 compatibility issues. Then, after some thought on how to implement this in a way that exposes entities to userland pyscripts, this temporary fix can be adjusted to fit in line with that plan.

stefanuytterhoeven:
Made the changes to my state.py, and can confirm it works...

craigbarratt:
@dlashua - thanks for the patch!  Couple of questions:

- is this backward compatible with older versions of HA?
- are there other breaking changes in HA 2022.3?

dlashua:
> * is this backward compatible with older versions of HA?

I don't think so. `async_restore_entity_added` was modified to expect an object passed to it. It then accesses attributes of that object. Is there a way to detect HA version and use different code depending on the version?

> * are there other breaking changes in HA 2022.3?

According to the release notes, no. But, the issue fixed here is also not mentioned in the release notes, so maybe?

I've not upgraded to 2022.3 yet, due to a lack of time. I just recently upgraded to 2022.2.

dlashua:
Oh my... all tests finally passed.

craigbarratt:
I'm not sure how to detect the HA version, although hopefully it is easy.  One approach would be to use try/except clauses to handle the new and old cases (which has the advantage that we don't need to explicitly test for the HA version).

dlashua:
After I finished fixing all the test failures, this will no longer load in Home Assistant.

ERROR (MainThread) [homeassistant.setup] Setup failed for pyscript: Integration not found.

But the files are in the right place (the same place they were in when I was testing the code changes). I have debug logging turned on and it's not providing any additional information. So... I'm not sure how I broke whatever is now broken.

I even deleted everything, recloned Home Assistant, and recloned the Pyscript component and I still have the same issue.

I'm out of time for now, perhaps @craigbarratt can figure out what went wrong?

dlashua:
Apparently the devcontainer got corrupted somehow. I have resolved that issue and this loads fine now.

dlashua:
This is now compatible with HASS Versions < 2022.3.0. I tested with 2022.2.9, but unless really old versions used some other method of restoring state, this should work.

dlashua:
I am now successfully running this in my "production" Home Assistant on version 2022.2.9.

For those needing a quick fix until this issue is ironed out and merged, you can install the master branch of github.com/dlashua/pyscript. If you use HACS, you can add this repo as a custom repository (it will show up as `pyscript_dlashua`). Then remove `pyscript`, install `pyscript_dlashua` instead, and then restart Home Assistant (apparently, HACS will allow two repos to provide files in the same custom_component at the same time, so, don't install `pyscript_dlashua` before removing `pyscript` or unknown things could happen).

dlashua:
I am now successfully running HASS 2022.3.1 without issues.

craigbarratt:
Thanks for the fix

I'm going to upgrade some of the versions in `tests/requirements_test.txt`.

In particular, `pytest-homeassistant-custom-component` v0.7.0 is based on 2022.3.1 so the next step is to make sure all the tests run with that version.

PR #373, opened 2022-07-23
Corrects the href back to HA's Dev Docs

robertgresock:
There was an extra hash at the beginning of the HREF tag. The hash acted as an HTML anchor on the current page.


craigbarratt:
Thanks for the fix!

PR #382, opened 2022-08-15
Fixes #374

### LINKED ISSUE #374: When state_hold != None then kwargs will not be set
pyscript Version 1.3.3
homeassistant Version 2022.6.7 on hassio on Raspberry Pi

When i set state_hold to a value != None, then kwargs will not be set correctly in a @state_trigger.

Correct: The following state_trigger with state_hold=None set "status=min" on triggering:
```@state_trigger(f"float({t_sensor}) < float({i_min_day})", state_hold_false=0, state_hold=None, state_check_now=True, kwargs={'status': 'min'})``` 

Wrong: The following state_trigger with state_hold=60 set "status=None" on triggering:
```@state_trigger(f"float({t_sensor}) < float({i_min_day})", state_hold_false=0, state_hold=60, state_check_now=True, kwargs={'status': 'min'})```

### PR DESCRIPTION
**Pull request by ALERTua**
short test to prove the fix works:
```
@state_trigger("input_boolean.test_boolean", state_hold=1, kwargs={'test_kwarg': 123})
def tryouts(trigger_type=None, var_name=None, value=None, old_value=None, context=None, **kwargs):
    log.debug(f"tryouts trigger_type: {trigger_type}, {kwargs}")
    if kwargs.get('test_kwarg') == 123:
        log.debug("after the fix")
    else:
        log.debug("before the fix")
```

### DIFF
--- a/custom_components/pyscript/trigger.py
+++ b/custom_components/pyscript/trigger.py
@@ -1116,6 +1116,7 @@ class TrigInfo:
                                     self.name,
                                     self.state_hold,
                                 )
+                            func_args.update(user_kwargs)
                             continue
                         if state_trig_waiting:
                             state_trig_waiting = False


craigbarratt:
Thanks for the fix!

PR #495, opened 2023-07-22
Added support for service responses when calling or creating services.

matzman666:
# Added support for returning service responses for both service execution and service creating.

## Calling Services:

- Added argument "return_response" to be passed to the call to the underlying home assistant service API.
- Added "return" to several places so that executed service calls can actually return values.
- Added some convenience functionality so that the user does not need to explicitly specify arguments "return_response" or "blocking" when calling a service which returns a response.

The user can now call services and get responses back. This works for both ways of calling services, via service.call() and also when using the service name as a function. Users can now use the argument "return_response = True" when they are interested in getting a response like this:

`retval = service.call("mydomain", "myservice", return_response = True, ...)`

or

`retval = mydomain.myservice(return_response = True, ...)`

When the called service has been created with "SupportsResponse.ONLY", then "return_response = True" is automatically assumed when not explicitly given by the user.

## Creating service:
- Decorator @service now accepts a new keyword argument "supports_response" which has the same semantics as the argument of the same name of the underlying home assistant API. When not explicitly given it has a default value of "none".
- The wrapper function of the custom service implementations now awaits the task and returns the return value.

The user can now use the keyword argument "supports_response" to specify whether the service returns a response when creating a service using the @service decorator:

```
@service(supports_response = "only")
def myservice():
    return {"response_data1": "somedata1", "response_data2": "somedata2"}

```
 The argument "supports_response" accepts the same values as the underlying home assistant API: "none" when no response is returned, "optional" when optionally a response can be returned or "only" when a response must be returned. Default value of "supports_response" is "none".

### DIFF
--- a/custom_components/pyscript/eval.py
+++ b/custom_components/pyscript/eval.py
@@ -15,6 +15,7 @@ import weakref
 
 import yaml
 
+from homeassistant.core import SupportsResponse
 from homeassistant.const import SERVICE_RELOAD
 from homeassistant.helpers.service import async_set_service_schema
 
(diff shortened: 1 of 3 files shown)


craigbarratt:
Thanks for the PR!  I'll make it backward compatible and add tests and documentation.

PR #500, opened 2023-07-27
Fix import of Event

rccoleman:
The Event class should have been imported from `homeassistant.core` (as per Frenck), and the current import breaks in 2023.8.  This change makes PyScript work properly in 2023.8.


craigbarratt:
Thanks for the fix!

PR #535, opened 2023-10-15
Ignore failure to delete list comp variables

### LINKED ISSUE #236: List comprehension failing with an empty list
The following very simple list comprehension works fine in the interactive Python interpreter, but fails if used in PyScript:

`[x for x in []]`

I would expect it to return an empty list (`[]`), as it does in the interactive interpreter:

```
>>> [x for x in []]
[]
```

but PyScript gives an error:

```
        [x for x in []]
                    ^
KeyError: 'x'
```

It's not a very interesting example, but I've had situations where an incoming iterable may be empty and it's easier to just execute the list comprehension without having to add an `if xxx` to the end to filter empty lists.  I've worked around it with a `for` loop, but I didn't see a limitation explicitly called out in the docs and wanted to file an issue for it.

Weirdly, I have seen this construct work as expected sometimes, but I can get to to reliably fail if I write an app like this:

```
@time_trigger('startup')
def startup():
    [x for x in []]
```

akonradi:
When iterating over a generator expression `gen` in a list comprehension `[for x in gen]`, the local variable `x` holds the value of each generated expression. When the list comprehension is done being evaluated, the local variable `x` needs to be deleted. This was already being done. If, however, `gen` is empty, `x` is never defined, and so deleting it causes an error. This patch adds a check to ignore that error.

Fixes #236


craigbarratt:
Thanks for fixing this issues and submitting a PR.  There are two `pylint` errors; could you fix those please?
```
custom_components/pyscript/eval.py:1758:16: C0103: Variable name "e" doesn't conform to snake_case naming style (invalid-name)
custom_components/pyscript/eval.py:175[8](https://github.com/custom-components/pyscript/actions/runs/6523765383/job/17800734376?pr=535#step:5:9):16: W0612: Unused variable 'e' (unused-variable)
```

akonradi:
Whoops, fixed.

craigbarratt:
Awesome!  Thanks for the PR.  Somehow I missed the original issue.

PR #592, opened 2024-03-12
Webhook Trigger

### LINKED ISSUE #353: Feature Request: Webhook support
I dont know if this is easily achievable, but i think it would provide useful, if functions could be triggered via a webhook, just like native automations.

Sometimes the data received via a webhook needs to be processed in some way that is easier in python as compared to an automation. Unfortunately webhooks themselves dont emit an event on which i could trigger.

My current workaround is an automation triggered via webhook, which then calls a service defined in pyscript. However, I only having to define the webhook in pyscript would provide useful.

contagon:
This PR closes #353  and adds a new webhook trigger defined as,

```python
@webhook_trigger(webhook_id, str_expr=None, local_only=True, methods={"POST", "PUT"}, kwargs=None)
def foo(payload):
    pass
```

When looking into this, I was hoping that webhooks trigger events, but they unfortunately don't leading to this solution. I've been using this for about a week and it seems to be working stably.  I've added a few things to the tests to make sure things at least run. I'm not sure how to actually call a webhook in a test, but happy to do it if there is a way. 

I also added some documentation that hopefully covers everything.

Let me know if any changes or tweaks are needed!

### DIFF
--- a/custom_components/pyscript/__init__.py
+++ b/custom_components/pyscript/__init__.py
@@ -51,6 +51,7 @@ from .mqtt import Mqtt
 from .requirements import install_requirements
 from .state import State, StateVal
 from .trigger import TrigTime
+from .webhook import Webhook
 
 _LOGGER = logging.getLogger(LOGGER_PATH)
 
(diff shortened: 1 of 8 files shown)


craigbarratt:
Awesome - thanks for the PR.  Sorry about the delay in following up.

PR #615, opened 2024-07-08
docs: Build out HACS installation instructions

amckenny:
This commit primarily builds out the HACS instructions on the installation doc page so that a new user who may not be familiar with HACS or have it installed will have enough background to get pyscript installed. Along with this, it synchronizes instructions regarding what to click with the button labels in Home Assistant (e.g., "Configuration ->" becomes "Settings -> Devices & services ->").

Secondary changes made address a few typographical issues in several documents. Only one change was made to a code snippet in 'tutorial.rst', relabeling 'pararm2' to 'param2'.

Note: At the end of 'tutorial.rst' there was a transition paragraph intended to introduce another section that would describe decorators/built-ins. This section was missing from the page, so I removed the transition paragraph. However, it could be added back when that section is finished.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -1,11 +1,25 @@
 Installation
 ============
 
-Option 1: HACS
+Option 1: Home Assistant Community Store (HACS)
 --------------
 
-Under HACS -> Integrations, select “+”, search for ``pyscript`` and
-install it.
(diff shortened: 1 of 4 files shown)


craigbarratt:
Thanks for the PR!

PR #634, opened 2024-09-10
adds state.exist classmethod to check whether an entity_id exists

ALERTua:
adds `state.exist` classmethod to check whether an entity_id exists
very handy to check things exist before manipulating them to get rid of the exceptions


craigbarratt:
Thanks for the PR!

PR #651, opened 2024-10-30
Fix import of Config for HA 2024.11.0b0

rccoleman:
The Config class moved to homeassistant.core_config in 2024.11.0b0.

### DIFF
--- a/custom_components/pyscript/__init__.py
+++ b/custom_components/pyscript/__init__.py
@@ -22,7 +22,7 @@ from homeassistant.const import (
     SERVICE_RELOAD,
 )
 from homeassistant.core import Event as HAEvent, HomeAssistant, ServiceCall
-from homeassistant.core_config import Config
+from homeassistant.helpers.typing import ConfigType
 from homeassistant.exceptions import HomeAssistantError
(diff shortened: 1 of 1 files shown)


custom_components/pyscript/__init__.py
```diff
@@ -21,7 +21,8 @@
     EVENT_STATE_CHANGED,
     SERVICE_RELOAD,
 )
-from homeassistant.core import Config, Event as HAEvent, HomeAssistant, ServiceCall
+from homeassistant.core import Event as HAEvent, HomeAssistant, ServiceCall
+from homeassistant.core_config import Config
```
MartinHjelmare, line 25: The second parameter to `async_setup` isn't an instance of `Config`. Please use `dict[str, Any]` or `ConfigType` instead.

MartinHjelmare (approved):
:+1:

custom_components/pyscript/__init__.py
rccoleman: Good call, changed and tested locally.  PR tests now pass as well.

ALERTua:
is this fix reverse-compatible? will I be able to use it on the stable 2024.10.4?

PR #680, opened 2025-01-13
Cosmetic: make @mqtt_trigger example ids consistent

larsimmisch:
The previous naming scheme was inconsistent.


craigbarratt:
Thanks for the PR!

PR #702, opened 2025-02-25
Attempt to fix config_entry warning

LucidityCrash:
I looked at https://developers.home-assistant.io/blog/2024/11/12/options-flow/
and made the suggested changes.  Not sure what to do with the self._show_form = False on line 30 though.

### DIFF
--- a/custom_components/pyscript/config_flow.py
+++ b/custom_components/pyscript/config_flow.py
@@ -24,9 +24,9 @@ PYSCRIPT_SCHEMA = vol.Schema(
 class PyscriptOptionsConfigFlow(config_entries.OptionsFlow):
     """Handle a pyscript options flow."""
 
-    def __init__(self, config_entry: ConfigEntry) -> None:
+    def __init__(self) -> None:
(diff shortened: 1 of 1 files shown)


LucidityCrash:
> Have you confirmed that this works correctly in HA / HACS (ie, the workflow for web-based settings works correctly)?
> 
> Unfortunately I didn't write the config flow code in pyscript, so I'm not sure what the right solution is.

I'm not 100% sure either ... Testing just carried out : 

1. I'd configured it using the config.yaml so I've now removed the 2 entries from that and then removed the integration.
2. Added the integraton - it and it gave me a dialogue box with the 2 tickboxes in.  I ticked them both as per my config.yaml (though I dont have any scripts that access hass variables). 
3. Executed one of my scripts that uses aiohttp and that worked.
4. Restarted Home assistant and re did 3.
5. Removed integration, restarted HA, repeated 2 & 3

All looks good to me.

craigbarratt:
Awesome - thanks for confirming.

PR #707, opened 2025-03-10
docs/reference.rst @mqtt_trigger and callbacks

dfries:
I just got bit by a couple things and wanted to contribute back to improve the documentation.
I think they are all minor changes or formatting and straight forward.

The biggest hurdle I had was trying to query the history, I was wanting yesterday's value, and everything in Home Assistant is geared around the most recent value.  In my case I was monitoring my car's TPMS and was wanting the leak rate, so now - yesterday.  I tried the Derivative sensor, but it does weighted averages for the time period, and since there's a large block of time each day that the car isn't available to get the pressure completely throws off the average leak rate.

```
instance = get_instance(hass)
future = await instance.async_add_executor_job(history.get_significant_states, hass, start, stop, ids)
```
I see the asyncio Futures page list `print(await fut)` for getting the future value, but that doesn't seem to work.  The best I've done is `future.add_done_callback` with all the drawbacks of needing to pass a compiled python function or lambda, it does work for what I need.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -774,15 +774,16 @@ on that topic. Multiple ``@mqtt_trigger`` decorators can be applied to a single 
 to trigger off different mqtt topics.
 
 An optional ``str_expr`` can be used to match the MQTT message data, and the trigger will only occur
-if that expression evaluates to ``True`` or non-zero. This expression has available these four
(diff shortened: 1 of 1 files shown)


craigbarratt:
Thanks for the doc updates!

PR #743, opened 2025-08-01
added last_reported to state attributes

Michael-CGN:
Add last_reported to the attributes of a state variable. The last_reported value was introduced to Homeassistant in 2024 to reflect that a value has been delivered to a state object. The existing last_updated and last_changed did not reflect this as they only show the last time the value really changed or another attribute changed


craigbarratt:
Thanks for the PR.  Can you also update the documentation?

craigbarratt:
Thanks for the PR.

Michael-CGN:
Thank you for this great custom component without which Homeassistant would be completely usesless for me :-)

PR #751, opened 2025-08-24
Do not defer listening to state_changed events until HA has started. …

### LINKED ISSUE #750: state_trigger not working after disabling and reactivating pyscript
I believe the culprit is here in `__init__.py`:
```
    async def hass_started(event: HAEvent) -> None:
        _LOGGER.debug("adding state changed listener and starting global contexts")
        await State.get_service_params()
        hass.data[DOMAIN][UNSUB_LISTENERS].append(hass.bus.async_listen(EVENT_STATE_CHANGED, state_changed))
        start_global_contexts()
```

rumpeltux:
Fixes #750 (see that issue for details)

### DIFF
--- a/custom_components/pyscript/__init__.py
+++ b/custom_components/pyscript/__init__.py
@@ -349,9 +349,8 @@ async def async_setup_entry(hass: HomeAssistant, config_entry: ConfigEntry) -> b
         await State.update(new_vars, func_args)
 
     async def hass_started(event: HAEvent) -> None:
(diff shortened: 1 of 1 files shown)


ALERTua:
@craigbarratt This PR claims to fix "state_trigger not working after disabling and reactivating pyscript":
The `state_changed` listener is only setup upon `hass_started`.
When disabling pyscript, the listener is removed. When re-enabling it, the listener is never re-added.

PR #771, opened 2025-10-15
Add IDE stubs generator service, docs, and tests

dmamelin:
This PR is a reworked and improved version of [pyscript_autocomplete](https://github.com/dmamelin/pyscript_autocomplete).
Better docstrings and more accurate type definitions for the Pyscript runtime and for Home Assistant domains, services, and entities with attributes.

How it works:
- Entity collection: `pyscript.generate_stubs` reads the HA entity registry, skips disabled entries, and records each entity along with its attributes. Attributes with names that aren’t valid Python identifiers are listed and excluded from the final code.
- Entity class generation: For each domain, a `_domain_state` class (subclass of `StateVal`) is created with the discovered attributes. Then a domain class is built where each entity is declared as a field of that state class.
- Service collection: Using `async_get_all_descriptions`, all HA service definitions are extracted. The generator converts selectors into python types, handles required/default fields and descriptions, and generates functions both at the domain level (as `@staticmethod`) and on the entity state class (if the service supports a target).
- AST-based code generation: Classes, declarations, and functions are constructed using the Python AST to ensure valid syntax, then converted to source with `ast.unparse`.
- File writing: The generator copies the bundled `pyscript_builtins.py` and writes the generated module to <config>/pyscript/modules/stubs/.
- Ignored imports: `from stubs import` and `from stubs.* import` statements are ignored at runtime. Theoretically this could be a breaking change, but in practice, no one keeps real code in stubs.

Example for `time` domain:
```python
class _time_state(StateVal):

    def set_value(self, time: str):
        """Sets the time.

        Args:
            time: The time to set. Example: 22:15"""
        ...

class time:
    time_settings_test: _time_state

    @staticmethod
    def set_value(*, entity_id: str, time: str):
        """Sets the time.

        Args:
            entity_id: Entity ID
            time: The time to set. Example: 22:15"""
        ...
```
In my HA installation, the generator produced about 5,900 lines of code, and it works perfectly.
Tested with PyCharm, VS Code, and Studio Code Server.
pycharm

### DIFF
--- a/custom_components/pyscript/__init__.py
+++ b/custom_components/pyscript/__init__.py
@@ -314,11 +314,8 @@ async def async_setup_entry(hass: HomeAssistant, config_entry: ConfigEntry) -> b
         def write_stubs(path) -> dict[str, Any]:
             res: dict[str, Any] = {}
             try:
-                if not os.access(path, os.W_OK):
-                    res["status"] = "Error"
-                    res["message"] = f" '{path}' is not writable."
-                    return res
                 os.makedirs(path, exist_ok=True)
+
                 builtins_path = os.path.join(os.path.dirname(__file__), "stubs", "pyscript_builtins.py")
                 shutil.copy2(builtins_path, path)
 


ALERTua:
`stubs` directory creation error was fixed with https://github.com/custom-components/pyscript/pull/771/commits/75769e3e4a414d43cfa7c879c6476f93c5c1b1f7
I have found no more problems.

craigbarratt:
Wow, this is impressive!  Thanks for the contribution.  Still reading through the code, and so far looks great.  I'll definitely merge this - hopefully tomorrow afternoon.

Issue #787, opened 2025-12-10
enum methods are not working


karabays:
Hi, 
I'm trying to use the Enum as it's introduced in the latest release but instance methods are not working as usual.

pyscript version: 1.7.0
HA version: 2025.12.2, home assistant OS

A simple class for home mode and a method to update entity state in HA.

```
class HomeState(Enum):
    HOME = "home"
    AWAY = "away"
    EXTENDED_AWAY = "extended_away"
    SLEEPING = "sleeping"
    JUST_ARRIVED = "just_arrived"

    def set_state(self):
        state.set("pyscript.home_state",self.value, new_attributes={"test":True})

state.persist("pyscript.home_state", default_value = HomeState.HOME.value,
                default_attributes = {"icon":"mdi:home-account"})

@time_trigger
def test():
    ha_state = HomeState.AWAY
    ha_state.set_state()
```
above the test function fails with the following error log:

```
This error originated from a custom integration.

Logger: custom_components.pyscript.file.location.test
Source: custom_components/pyscript/trigger.py:1397
integration: Pyscript Python scripting (documentation, issues)
First occurred: 14:29:57 (1 occurrence)
Last logged: 14:29:57

Exception in <file.location.test> line 36: ha_state.set_state() ^ TypeError: 'set_state' is not callable (got HomeState.set_state) Traceback (most recent call last): File "/config/custom_components/pyscript/eval.py", line 946, in aeval val = await getattr(self, name, self.ast_not_implemented)(arg) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "/config/custom_components/pyscript/eval.py", line 1967, in ast_call return await self.call_func(func, func_name, *args, **kwargs) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "/config/custom_components/pyscript/eval.py", line 2019, in call_func raise TypeError(f"'{func_name}' is not callable (got {func})") TypeError: 'set_state' is not callable (got HomeState.set_state)
```
but this works:

```
def set_state_in_ha(home_state:HomeState):
    state.set("pyscript.home_state",home_state.value)
```

Am I doing something wrong or `enum` is not fully functional in pyscript.

PR #790, opened 2025-12-11
Add EvalFuncVar binding via descriptor and drop manual rebinding

dmamelin:
Looking into #787 I found that `Enum` overrides `__dir__`, so [dir(inst)](https://github.com/custom-components/pyscript/compare/master...dmamelin:pyscript:evalfuncvar-descriptor-binding?expand=1#diff-733b200d9ce0361a3582c281043fd548b057e971f927f73aed75bfdc8422f06cL1992) inside `call_func` doesn’t show the method.
I went deep into MRO, inspect, etc., but the actual fix turned out to be much simpler: [descriptors](https://docs.python.org/3/howto/descriptor.html) solve this problem.
[CPython implementation](https://github.com/python/cpython/blob/main/Objects/funcobject.c#L1199)

A quick search shows that `__dir__` is rarely overridden, but this PR may also affect other class-related issues in Pyscript - I just don’t know which ones yet :)

The logic for creating `EvalFuncVarClassInst` has also changed. Previously, an `EvalFuncVarClassInst` was created for every available `EvalFuncVar` method when the instance was constructed.
Now, no `EvalFuncVarClassInst` objects are created during instance construction; instead, a new `EvalFuncVarClassInst` is created on each access to an EvalFuncVar.

```python
class Test:
    def a(self):
        pass

    def b(self):
        pass

def old():
    test = Test()  # AstEval.call_func: create EvalFuncVarClassInst for a() and b()
    test.a()       # use existing EvalFuncVarClassInst
    test.a()       # use existing EvalFuncVarClassInst

def new():
    test = Test()  # AstEval.call_func: _NO_ EvalFuncVarClassInst created
    test.a()       # create EvalFuncVarClassInst and call it
    test.a()       # create a new EvalFuncVarClassInst and call it
    b = test.a     # create EvalFuncVarClassInst
    b()            # call the existing EvalFuncVarClassInst
```

### DIFF
--- a/custom_components/pyscript/eval.py
+++ b/custom_components/pyscript/eval.py
@@ -873,6 +873,15 @@ class EvalFuncVar:
         """Return the ast context."""
         return self.ast_ctx
 
+    def __get__(self, obj, objtype=None):
+        """Support descriptor protocol so class attributes bind to instances."""
+        if obj is None:
+            return self
+        # we use weak references when we bind the method calls to the instance inst;
+        # otherwise these self references cause the object to not be deleted until
+        # it is later garbage collected
+        return EvalFuncVarClassInst(self.func, self.ast_ctx, weakref.ref(obj))
+
     def __del__(self):
         """On deletion, stop any triggers for this function."""
         if self.func:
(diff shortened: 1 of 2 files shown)


dmamelin:
The test failure is not related to the modified code.

dmamelin:
https://github.com/aio-libs/aiodns/issues/214

Issue #820, opened 2026-04-14
Intermittent malformed payload for mqtt_trigger


ALERTua:
```
2026-04-13 12:45:22.569 ERROR (MainThread) [custom_components.pyscript.file.xiaomi_wireless_buttons] 
  File "/config/custom_components/pyscript/decorators/base.py", line 56, in check_expression_vars
    return await self._ast_expression.eval(state_vars)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/pyscript/xiaomi_wireless_buttons.py", line 1, in file.xiaomi_wireless_buttons.button_2_single @mqtt_trigger()
    payload_obj['action'] == 'single'
    ~~~~~~~~~~~^^^^^^^^^^
KeyError: 'action'
```

triggered for

```python
@mqtt_trigger('zigbee2mqtt_slzb/xiaomi_wireless_button_2', "payload_obj['action'] == 'single'")
def button_2_single(trigger_type=None, var_name=None, value=None, old_value=None, context=None, **kwargs):
    log.debug(f"button_2_single")
```

I cannot reproduce this 100% of the presses. Maybe the button randomly sends malformed payloads. Disregard until an issue is raised, I guess.
I changed it to `payload_obj.get('action', None)`.

I could not catch the event to get the real contents of the payload, as I don't understand how to do it in the UI of HA. I'm stupid or the events subscription just does not work.

via https://github.com/custom-components/pyscript/pull/818
ping @dmamelin

dmamelin:
Before #813, errors in mqtt_trigger expressions - and possibly in all trigger expressions - were completely ignored. I'm working on a fix that preserves backward compatibility and provides proper error reporting.
Thanks for catching this!

dmamelin:
In DM, everything works as expected. If an error occurs, the exception is logged and the triggers continue to work normally (previously, such errors were hidden from the user).
A bug was also found in the legacy implementation - a PR is coming soon.

> I could not catch the event to get the real contents of the payload, as I don't understand how to do it in the UI of HA. I'm stupid or the events subscription just does not work.

Agreed - too much is hidden from debugging right now. I have some ideas to improve the UX around decorators, but I'll look into them after the DM release.

PR #823, opened 2026-04-15
Fix regression in legacy trigger expressions

dmamelin:
Based on #820:

`@event_trigger`, `@mqtt_trigger`, and `@webhook_trigger` had been completely hiding expression errors for a long time.
After #813, triggers in the legacy implementation started failing after the first error.

`@state_trigger` was not affected by #813, because it already had error handling from the start.
DM is also not affected due to its architecture


craigbarratt:
Thanks for the PR

PR #825, opened 2026-04-16
Add webhook trigger decorator stub

nolsto:
Webhook trigger decorator was left out of stubs generation. This PR adds a builtin stub for it, and also updates some outdated language in the documentation.


craigbarratt:
Thanks for the PR.  Can you fix the minor formatting issue please?

nolsto:
Thanks! I should've run the pre-commit hook.
I also disabled a pylint rule (for mutable set as default param value) in the stubs module. Seems fine to me, but let me know if that is a problem.

Issue #827, opened 2026-04-29
2026.5 not working


Wesley-Vos:
Hi,

I've upgraded to 2026.5.0b0 but now the integration doesn't work anymore. It tries to import a helper that doesn't exist anymore or has been moved.

The log is as follows:
```

Logger: homeassistant.setup
Bron: setup.py:343
Eerst voorgekomen: 19:25:39 (1 gebeurtenis)
Laatst gelogd: 19:25:39

Setup failed for custom integration 'pyscript': Unable to import component: cannot import name 'forgiving_boolean' from 'homeassistant.helpers.template' (/usr/src/homeassistant/homeassistant/helpers/template/__init__.py)
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/loader.py", line 1024, in async_get_component
    comp = await self.hass.async_add_import_executor_job(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        self._get_component, True
        ^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/local/lib/python3.14/concurrent/futures/thread.py", line 86, in run
    result = ctx.run(self.task)
  File "/usr/local/lib/python3.14/concurrent/futures/thread.py", line 73, in run
    return fn(*args, **kwargs)
  File "/usr/src/homeassistant/homeassistant/loader.py", line 1084, in _get_component
    ComponentProtocol, importlib.import_module(self.pkg_path)
                       ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/util/loop.py", line 201, in protected_loop_func
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.14/importlib/__init__.py", line 88, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1398, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1371, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1342, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 938, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 759, in exec_module
[...]

PR #818, opened 2026-04-12
Decorator Manager

dmamelin:
Clean PR with a single commit. Full description, discussion, and commit history are in #795.

PR #828, opened 2026-05-01
HA 2026.5 support

dmamelin:
Fixing #827 required updating `pytest-homeassistant-custom-component`, which in turn requires Python 3.14.
But after moving to the new Python version, the tests broke: HA could no longer load the pyscript integration because `os.path.isfile` was being patched globally in the tests.

I don't know why this worked before - `homeassistant/loader.py` hasn't changed in a long time. Most likely this is due to changes in CPython: https://github.com/python/cpython/pull/118243

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -30,7 +30,7 @@ jobs:
       - uses: "actions/checkout@v2"
       - uses: "actions/setup-python@v1"
         with:
-          python-version: "3.13"
+          python-version: "3.14"
           allow-prereleases: true
       - run: python3 -m pip install -r tests/requirements_test.txt
       - run: pytest --cov=custom_components
@@ -42,7 +42,7 @@ jobs:
       - uses: "actions/checkout@v2"
       - uses: "actions/setup-python@v1"
         with:
-          python-version: "3.13"
+          python-version: "3.14"
           allow-prereleases: true
       - run: python3 -m pip install -r tests/requirements_test.txt
       - run: pylint custom_components/pyscript/*.py tests/*.py
@@ -54,7 +54,7 @@ jobs:
       - uses: "actions/checkout@v2"
       - uses: "actions/setup-python@v1"
         with:
-          python-version: "3.13"
+          python-version: "3.14"
           allow-prereleases: true
       - run: python3 -m pip install -r tests/requirements_test.txt
       - run: mypy custom_components/pyscript/*.py tests/*.py
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.14
+    rev: v0.15.12
     hooks:
       - id: ruff-check
         args: [--fix, --show-fixes]
--- a/custom_components/pyscript/__init__.py
+++ b/custom_components/pyscript/__init__.py
@@ -6,6 +6,8 @@ import glob
 import json
 import logging
 import os
+from os import makedirs as os_makedirs
+from os.path import isdir as os_path_isdir
 import shutil
 import time
 import traceback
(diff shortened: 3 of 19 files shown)


dmamelin:
A small dependency-version hell - everything works locally, but I need a bit more time.

dmamelin:
I pinned the minimal version to 2026.5.0 expecting the usual May 1 release, but it's now scheduled for May 6.

rccoleman:
> I pinned the minimal version to 2026.5.0 expecting the usual May 1 release, but it's now scheduled for May 6.

HA releases are always on the first Wed of the month.  Betas start on the last Wed of the month.

dmamelin:
My mistake. However, I suggest leaving this unchanged and keeping Home Assistant 2026.5.0 as the requirement for Pyscript 2.0.

For regular users, the HA and Pyscript updates will be released at the same time.

craigbarratt:
@dmamelin thanks for the fixes.  Just wondering what the most prudent release strategy is

1. do a stepping-stone release with these fixes for 2026.5.0 without #818
2. release 2.0.0 now with all PRs and fixes for 2026.5.0

I think the 2nd is the best option, although we will be front-loading issues on top of the 2026.5.0 changes since there won't be any fallback with 2026.5.0 support, nor a beta-test period with #818.

Also, how important is it for 2.0.0 to be backward compatible with HA < 2026.5.0?  I don't this PR will be backward compatible.

Thoughts?

dmamelin:
@craigbarratt i prefer the second option. There's no point in excluding #818, because the `legacy_decorators` option exists.
BTW, there may still be bugs in the recent changes, but I'll be fixing them quickly.

craigbarratt:
Ok, great, I agree - forward progress without orphan releases.  I'm going to update the imports in this PR to hopefully make it backward compatible (with try / except on the imports).  See below.

Also, there were changes from "except (ValueError, TypeError)" to "except ValueError, TypeError".  I believe the former is correct, so I'll change them back.

```
try:
    from homeassistant.helpers.template.extensions.math import _SENTINEL as MATH_SENTINEL, MathExtension
    from homeassistant.helpers.template.extensions.type_cast import (
        TypeCastExtension,
        _SENTINEL as TYPECAST_SENTINEL,
    )
    from homeassistant.helpers.template.helpers import (
        _SENTINEL,
        forgiving_boolean,
        raise_no_default,
    )
except ImportError:
    from homeassistant.helpers.template import (  # type: ignore[no-redef]
        _SENTINEL,
        forgiving_boolean,
        forgiving_float as _forgiving_float,
        forgiving_int as _forgiving_int,
        forgiving_round as _forgiving_round,
        raise_no_default,
    )

    TYPECAST_SENTINEL = _SENTINEL
    MATH_SENTINEL = _SENTINEL

    class TypeCastExtension:  # type: ignore[no-redef]
        """Shim for older HA versions that export forgiving_* as module-level functions."""

        @staticmethod
        def forgiving_float(value, default=_SENTINEL):
            return _forgiving_float(value, default=default)

        @staticmethod
        def forgiving_int(value, default=_SENTINEL, base=10):
            return _forgiving_int(value, default=default, base=base)

    class MathExtension:  # type: ignore[no-redef]
        """Shim for older HA versions that export forgiving_round as a module-level function."""

        @staticmethod
        def forgiving_round(value, precision=0, method="common", default=_SENTINEL):
            return _forgiving_round(value, precision=precision, method=method, default=default)
```

dmamelin:
> Also, there were changes from "except (ValueError, TypeError)" to "except ValueError, TypeError". I believe the former is correct, so I'll change them back.

@craigbarratt this is required by the new `ruff format`. Reverting to the previous style and ignoring the new PEP would mean sticking with an older ruff version.

The idea of using try / except on the imports is excellent.

craigbarratt:
@dmamelin good point about PEP 758.  However, it's a syntax error in python <= 3.13.  I'd rather keep backward compatibility for a few months, and we'll need to pin an older ruff version.

dmamelin:
I'll try to be more mindful of backward compatibility :)

My reasoning was simple: if a user updates integrations through HACS, Home Assistant itself is usually up to date as well. And Python >= 3.14.2 is required for Home Assistant starting from 2026.3.0

craigbarratt:
@dmamelin Ok, this should be backward compatible.  Anything else to do or check before releasing 2.0.0?

dmamelin:
@craigbarratt i don't know the full release process, but from the code side everything is ready.

craigbarratt:
2.0.0 released!

dmamelin:
@craigbarratt thanks! Hopefully the release is boring in the best possible way :)

Issue #836, opened 2026-05-05
Startup error in 2.0.0


attila123456:
After updating to release 2.0.0 and restarting HA, I get several errors in the log:

```
ERROR (MainThread) [custom_components.pyscript.file.lights] FunctionDecoratorManager(running) [@webhook_trigger(...), @time_active(...)] for file.lights...> start failed: Handler is already defined!
Traceback (most recent call last):
  File "/config/custom_components/pyscript/decorator_abc.py", line 180, in start
    await decorator.start()
  File "/config/custom_components/pyscript/decorators/webhook.py", line 69, in start
    webhook.async_register(
    ~~~~~~~~~~~~~~~~~~~~~~^
        self.dm.hass,
        ^^^^^^^^^^^^^
    ...<5 lines>...
        allowed_methods=self.methods,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/src/homeassistant/homeassistant/components/webhook/__init__.py", line 54, in async_register
    raise ValueError("Handler is already defined!")
ValueError: Handler is already defined!
```

Enabling legacy decorators in the options does resolve the issue.

dmamelin:
Found a difference between the legacy and new implementations. Are you using `@webhook_trigger` with the same `webhook_id` in multiple places?

attila123456:
Yes, a "person-detected" webhook, called by unifi cameras, used as a trigger in multiple pyscript automations (lights, alarm).

dmamelin:
Version 2.0.1 has been released with the fix.

attila123456:
Thanks, no errors anymore!

Issue #834, opened 2026-05-05
NameError: name 'now' is not defined


shashchatter:
This is a new error, got on first HA restart after updating pyScript to 2.0.0 and HA to 2026.4.4.

2026-05-05 16:00:00.012 ERROR (MainThread) [custom_components.pyscript.function] run_coro: got exception Traceback (most recent call last):
  File "/config/custom_components/pyscript/eval.py", line 936, in aeval
    raise NameError(f"name '{val.name}' is not defined")
NameError: name 'now' is not defined

PyScript: 2.0.0

Installation method Home Assistant Container
Core 2026.4.4
Frontend 20260325.8

dmamelin:
It's not clear yet what the issue is. Please enable debug logging for Pyscript and restart Home Assistant.
Add `custom_components.pyscript: debug` to the logger.logs section in `configuration.yaml`:
```yaml
logger:
  default: info
  logs:
    custom_components.pyscript: debug
```

craigbarratt:
Can you also include the `pyscript` user code is being executed, ideally localized to where the error occurs?  Prior versions of `pyscript` silently ignored certain errors.  It's quite possible there is a latent error in your user code that is now being reported.

shashchatter:
> It's not clear yet what the issue is. Please enable debug logging for Pyscript and restart Home Assistant. Add `custom_components.pyscript: debug` to the logger.logs section in `configuration.yaml`:
> 
> logger:
>   default: info
>   logs:
>     custom_components.pyscript: debug

I was unable to reproduce this with the logger set to debug, and then back to info.  I will be on the lookout to see if this happens again.

shashchatter:
> Can you also include the `pyscript` user code is being executed, ideally localized to where the error occurs? Prior versions of `pyscript` silently ignored certain errors. It's quite possible there is a latent error in your user code that is now being reported.

You are right, of course it could be latent in user code.  I haven't been able to reproduce the issue yet.  Will keep watching.  I wasn't running or invoking a script manually through a button or call from Settings->Developer tools.  It must be from code ececuted from one  of the triggers, not sure yet about which one.

shashchatter:
Lucked out, saw the error, with a second error with the function name, both errors turned out to be in the same function:

```
2026-05-06 20:00:00.023 ERROR (MainThread) [custom_components.pyscript.function] run_coro: got exception Traceback (most recent call last):
  File "/config/custom_components/pyscript/eval.py", line 936, in aeval
[...]

PR #837, opened 2026-05-06
DM fixes (detailed log & webhook)

dmamelin:
In #836, I noticed a difference in the `@webhook_trigger` implementation and aligned it with the legacy behavior.

In #834, this is definitely a bug in user code, but I also noticed that the exception message lacks useful detail.
While porting DM to the new error-handling subsystem, I introduced a bug - exceptions raised inside triggered functions become uninformative.

@craigbarratt, please release a hotfix.


craigbarratt:
Released 2.0.1

PR #842, opened 2026-05-10
expose aiohttp request to @webhook_trigger functions

JayNewstrom:
Add the underlying aiohttp.web.Request to the kwargs passed to @webhook_trigger user functions (and to the str_expr filter). Lets scripts inspect headers, method, query string, and re-read the raw body via `await request.read()` -- enabling HMAC signature validation, which previously wasn't possible since only the parsed payload was exposed.

Backward compatible: existing functions that don't declare `request` are unaffected.

### DIFF
--- a/custom_components/pyscript/decorators/webhook.py
+++ b/custom_components/pyscript/decorators/webhook.py
@@ -54,6 +54,7 @@ class WebhookTriggerDecorator(TriggerDecorator, ExpressionDecorator, AutoKwargsD
         func_args = {
             "trigger_type": "webhook",
             "webhook_id": webhook_id,
+            "request": request,
         }
 
         if "json" in request.headers.get(hdrs.CONTENT_TYPE, ""):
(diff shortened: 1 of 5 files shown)


craigbarratt:
Thanks for the PR.  Can you also update `custom_components/pyscript/stubs/pyscript_builtins.py`?

JayNewstrom:
Thanks for the review -- updated.

JayNewstrom:
The alternative here is to add `headers` as a map, and `raw_payload` to kwargs. Less flexible/powerful, but the targeted fields I'd need to do hmac validation for the webhooks. Happy to make the change if you'd prefer.

JayNewstrom:
I guess another alternative to solve my use case would be to add params to the webhook trigger annotation for the webhook secret and header name. Given those the validation could happen inside the pyscript code. But maybe that's too involved on the pyscript side.

craigbarratt:
Thanks for the PR!

PR #843, opened 2026-05-12
allow @webhook_trigger functions to control the http response

JayNewstrom:
Webhook handlers now wait for the decorated function and use its return value to build the response: int -> Response(status=...), aiohttp.web.Response -> as-is, None -> default 200.

### DIFF
--- a/custom_components/pyscript/decorator.py
+++ b/custom_components/pyscript/decorator.py
@@ -264,6 +264,7 @@ class FunctionDecoratorManager(DecoratorManager):
                 # notify handlers with "None"
                 for result_handler_dec in result_handlers:
                     await result_handler_dec.handle_call_result(data, None)
+                data.set_result(None)
                 return
         # Fire an event indicating that pyscript is running
         # Note: the event must have an entity_id for logbook to work correctly.
@@ -279,7 +280,9 @@ class FunctionDecoratorManager(DecoratorManager):
             result = await data.call_ast_ctx.call_func(self.eval_func, None, **data.func_args)
             for result_handler_dec in result_handlers:
                 await result_handler_dec.handle_call_result(data, result)
+            data.set_result(result)
         except Exception as e:
+            data.set_result(None)
             await self.handle_exception(e)
 
     async def dispatch(self, data: DispatchData) -> None:
@@ -290,6 +293,7 @@ class FunctionDecoratorManager(DecoratorManager):
         for dec in decorators:
             if await dec.handle_dispatch(data) is False:
                 self.logger.debug("Trigger not active due to %s", dec)
+                data.set_result(None)
                 return
 
         action_ast_ctx = AstEval(
--- a/custom_components/pyscript/decorator.py
+++ b/custom_components/pyscript/decorator.py
@@ -264,7 +264,6 @@ class FunctionDecoratorManager(DecoratorManager):
                 # notify handlers with "None"
                 for result_handler_dec in result_handlers:
                     await result_handler_dec.handle_call_result(data, None)
-                data.set_result(None)
                 return
         # Fire an event indicating that pyscript is running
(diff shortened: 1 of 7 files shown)


dmamelin:
For handling function call results, there is a `CallResultHandlerDecorator` class. It isn't used yet, but it was intended for cases like this.
```python
class WebhookTriggerDecorator(..., CallResultHandlerDecorator):
    ...
    
    async def handle_call_result(self, data: DispatchData, result) -> None:
        if data.trigger is not self:
            return
        response_future = data.trigger.future
        if response_future is not None and not response_future.done():
            response_future.set_result(result)
```

dmamelin:
Thanks for the changes! Everything now aligns with the new architecture.

Possibly, adding a `sets_http_response_code` flag is unnecessary - function results could simply always be handled. That would be more intuitive, would not break backward compatibility, and should not add much overhead.

But of course, the final decision is up to @craigbarratt.

JayNewstrom:
My motivation for sets_http_response_code was primarily to drive validations. Otherwise the behavior might be undefined if there are 2 webhook_triggers which return an HTTP response.

But happy to defer to you on preferences here. Just let me know and I can update the functionality.

dmamelin:
There is also another issue with two webhooks in the original implementation of this trigger:
```python
@webhook_trigger("test", methods=["GET"])
def a():
    ...

@webhook_trigger("test", methods=["POST"])
def b():
    ...
```
In this case, a GET request will call both functions, while a POST request will call neither.

dmamelin:
@craigbarratt , I would deprecate support for multiple webhooks with the same ID and remove it after a few releases. That would be more logical and would avoid conflicts with Home Assistant's internal implementation.

dmamelin:
As an option, `webhook_trigger` could remain a simple notification-style decorator: allowing duplicate IDs, with no `response` support and no `request` support.

Then a separate `webhook_handler` (or similar) could provide the extended behavior: unique IDs, support for the `request` argument, and always returning a `response`.

craigbarratt:
> There is also another issue with two webhooks in the original implementation of this trigger:
> 
> ```python
> @webhook_trigger("test", methods=["GET"])
> def a():
>     ...
> 
> @webhook_trigger("test", methods=["POST"])
> def b():
>     ...
> ```
> 
> In this case, a GET request will call both functions, while a POST request will call neither.

I agree that's a bug.  The webhook code was user submitted, so I'm not too familiar with it.

Back to the PR, does this require the webhook trigger function to return (ie, finish), before any http response is generated, in the case where `sets_http_response_code` is not set?  If so, that could be a breaking change.

I prefer @dmamelin's suggestion of not overloading `webhook_trigger` and adding a new `webhook_handler` with the extended behavior.  The duplicate ID issue should still be deprecated or fixed.

craigbarratt:
I asked claude code to review, and it suggests there could be race conditions with concurrent requests to the same webbook_id.  This further supports @dmamelin's suggestion that it should be a new trigger with unique IDs and new response semantics, with careful implementation to handle multiple overlapping web requests to the same trigger:
```
Trigger instances live in webhook_id2triggers and are reused for every request. Two concurrent requests to the same webhook_id race:

- R1 sets trigger.future = F1, dispatches (which only schedules _call as a task and returns).
- R2 sets trigger.future = F2, dispatches.
- _call for R1 runs, calls handle_call_result, reads self.future → now F2, resolves F2 with R1's result.
- _call for R2 reads self.future → F2 (already done), skips. F1 is never resolved.

Result: R1's asyncio.gather([F1]) hangs forever (the HTTP request never returns), and R2 receives R1's response. Concurrent webhook hits are realistic, so this isn't a corner case.
```

JayNewstrom:
I'm going to close this PR, and open 2 new PRs:
1. New webhook_handler annotation
2. Fix webhook_trigger to only allow 1

PR #844, opened 2026-05-13
Add hold_off_send_last to @time_active

dmamelin:
The existing `hold_off` drops all triggers during the hold-off window, which is useful for rate limiting but can lose the latest data.
With `hold_off_send_last` enabled, it still suppresses intermediate triggers, but runs once after the hold-off expires using the most recent suppressed trigger data.

### DIFF
--- a/custom_components/pyscript/decorators/timing.py
+++ b/custom_components/pyscript/decorators/timing.py
@@ -30,17 +30,48 @@ class TimeActiveDecorator(TriggerHandlerDecorator, AutoKwargsDecorator):
 
     name = "time_active"
     args_schema = vol.Schema(vol.All([vol.Coerce(str)], vol.Length(min=0)))
-    kwargs_schema = vol.Schema({vol.Optional("hold_off", default=0.0): vol.Any(None, cv.positive_float)})
+    kwargs_schema = vol.Schema(
+        {
+            vol.Optional("hold_off", default=0.0): vol.Any(None, cv.positive_float),
+            vol.Optional("hold_off_send_last", default=False): cv.boolean,
+        }
+    )
 
     hold_off: float | None
+    hold_off_send_last: bool
 
     last_trig_time: float = 0.0
+    _hold_off_task: asyncio.Task | None = None
+    _pending_data: DispatchData | None = None
+
+    async def _dispatch_after_hold_off(self) -> None:
+        """Dispatch the latest suppressed payload after the current hold-off window."""
+        while self._pending_data is not None:
+            delay = self.last_trig_time + self.hold_off - time.monotonic()
+            if delay > 0.0:
+                await asyncio.sleep(delay)
+
+            data = self._pending_data
+            _LOGGER.debug("%s hold_off_send_last dispatching after delay %s", self, delay)
+            await self.dm.dispatch(data)
+            if self._pending_data is data:
+                self._pending_data = None
 
     async def handle_dispatch(self, data: DispatchData) -> bool:
         """Handle dispatch."""
         if self.last_trig_time > 0.0 and self.hold_off is not None and self.hold_off > 0.0:
             if time.monotonic() - self.last_trig_time < self.hold_off:
+                if self.hold_off_send_last:
+                    self._pending_data = data
+                    if self._hold_off_task is None or self._hold_off_task.done():
+                        self._hold_off_task = self.dm.hass.async_create_background_task(
+                            self._dispatch_after_hold_off(), f"{self} hold_off_send_last"
+                        )
                 return False
+
+        if data is self._pending_data:
+            self.last_trig_time = time.monotonic()
+            return True
 
         if len(self.args) > 0:
             if "trigger_time" in data.func_args and isinstance(data.func_args["trigger_time"], dt.datetime):
@@ -53,11 +84,22 @@ class TimeActiveDecorator(TriggerHandlerDecorator, AutoKwargsDecorator):
                 _LOGGER.debug("time_active now %s, %s", now, self)
                 if await trigger.TrigTime.timer_active_check(time_spec, now, self.dm.startup_time):
                     self.last_trig_time = time.monotonic()
+                    if data is not self._pending_data:
+                        self._pending_data = None
                     return True
             return False
 
         self.last_trig_time = time.monotonic()
+        if data is not self._pending_data:
+            self._pending_data = None
         return True
+
+    async def stop(self) -> None:
+        """Stop pending hold-off dispatch."""
+        await super().stop()
+        self._pending_data = None
+        if self._hold_off_task is not None:
+            self._hold_off_task.cancel()
 
 
 class TimeTriggerDecorator(TriggerDecorator):
--- a/custom_components/pyscript/decorators/timing.py
+++ b/custom_components/pyscript/decorators/timing.py
@@ -49,10 +49,13 @@ class TimeActiveDecorator(TriggerHandlerDecorator, AutoKwargsDecorator):
         while self._pending_data is not None:
             delay = self.last_trig_time + self.hold_off - time.monotonic()
             if delay > 0.0:
+                # A trigger may pass and clear _pending_data while we sleep;
+                # re-check the loop condition after every await.
                 await asyncio.sleep(delay)
+                continue
 
             data = self._pending_data
-            _LOGGER.debug("%s hold_off_send_last dispatching after delay %s", self, delay)
+            _LOGGER.debug("%s hold_off_send_last dispatching %s", self, data)
             await self.dm.dispatch(data)
             if self._pending_data is data:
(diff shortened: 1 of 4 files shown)


ALERTua:
As a dyslexic, I would want to have an example for this kwarg, as, after reading the description thrice, I could not grasp its functionality. You may disregard my comment.

dmamelin:
Example: if `sensor.power` changes to `10`, then `20`, then `30` within 60 seconds:

  - With `@time_active(hold_off=60)`, the function runs immediately for `10`, and `20`/`30` are dropped.
  - With `@time_active(hold_off=60, hold_off_send_last=True)`, the function runs immediately for `10`, suppresses `20`, and then runs once more after 60 seconds with `30`.

This is useful for noisy and unstable sensors that spam updates: decorator can ignore the intermediate churn while still processing the latest value once the hold-off period ends.

ALERTua:
So `hold_off_send_last` holds for `hold_off` seconds after the first trigger, but then triggers again for the last trigger that was within those `hold_off` seconds?

dmamelin:
`hold_off_send_last` controls what happens to data dropped during `hold_off`.

With `hold_off_send_last=True`, the dropped data will be sent after `hold_off` seconds from the last send.
With `hold_off_send_last=False` (default), the dropped data will be ignored completely, as before.

There is no double send.

ALERTua:
So if my power voltage sensor trigger is `voltage >= 180` with `hold_off=60` and `hold_off_send_last=True`, the `voltage` was `220`, triggered on `0`, then, after a second, turned `unavailable`, then, after 15 seconds, back to `220`, the triggers would be `0`, and then, after the `hold_off` period, `220` (`unavailable` being skipped)?

dmamelin:
With `@state_trigger("voltage >= 180")`, if `voltage=0` or `voltage=unavailable`, execution will not even reach `@time_active`, because `@state_trigger` will not fire.

With `@state_trigger("voltage >= 0")` and `@time_active(hold_off=60, hold_off_send_last=True)`, it will behave as you described: `0` will be sent, `unavailable` will be dropped, and 220 will be delivered after `hold_off`.

ALERTua:
Crap, you are right. I made up a bad example.

dmamelin:
@ALERTua, i corrected my comment above.

If it's still unclear, I'm happy to walk through more examples.

It seems I didn't phrase the comment in `reference.rst` clearly enough. Let's try to find the right wording together - any suggestions?

ALERTua:
Thank you, now I understand the functionality behind `hold_off_send_last`. I asked Claude to find a good phrasing and Opus 4.7 suggested:
```
If `hold_off_send_last` is true, triggers that arrive during the `hold_off` window are still suppressed, but the function is run **once more** after the window ends, using the data from the **most recent** suppressed trigger. Earlier suppressed triggers are discarded. If no triggers arrive during the window, no extra run is scheduled.

This is useful for noisy sensors where you want to rate-limit execution but still react to the latest value once things settle.

Example:
with `@time_active(hold_off=60, hold_off_send_last=True)`, if `sensor.my_sensor` changes to `10`, then `20`, then `30` within 60 seconds, the function runs immediately for `10`, suppresses `20`, and runs once more after the hold-off ends with `30`. Without `hold_off_send_last`, only the run for `10` happens and `20`/`30` are dropped entirely.
```

craigbarratt:
Overall looks good, but a few issues raised by a claude code review:

1. The new deferred trigger could fall outside the `@time_active` window, since it isn't rechecked and it will always happen.  Perhaps that's the right behavior; `@time_active` is checked on the first trigger.  So perhaps the solution is to note that `@time_active` doesn't apply to the `hold_off_send_last` trigger in the docs.
2. There's a potential race condition at the very end of the hold off if a new trigger occurs (I haven't confirmed this, but here's what claude claims:

_dispatch_after_hold_off re-reads pending state after sleeping:
```
while self._pending_data is not None:
    delay = self.last_trig_time + self.hold_off - time.monotonic()
    if delay > 0.0:
        await asyncio.sleep(delay)
    data = self._pending_data        # could be None now
    await self.dm.dispatch(data)
```
The sleep ends exactly at last_trig_time + hold_off, which is the same instant a direct trigger first escapes the hold-off window and clears _pending_data = None (the success branches do self._pending_data = None). If that direct trigger's handle_dispatch runs first on that tick, data becomes None and self.dm.dispatch(None) runs. Any TriggerHandlerDecorator ordered before time_active that touches data.func_args (e.g. state_active) then hits AttributeError. It's a narrow tie, but cheap to close — re-check if data is None: break after the sleep, or snapshot before sleeping.

Also, in the new tests, test_time_active_hold_off_send_last uses hold_off=0.05 and relies on var1=3/var1=4 landing within 50 ms of the first trigger's last_trig_time.  Under load, that time could be exceeded.  Can the hold_off be longer or advancing time manually rather than relying on real-time clock?

dmamelin:
Good points.
I fully agree with the second one. I thought a lot about the first one: should the `@time_active` window be checked again after the `hold_off` window? I could not find a clearly correct solution.
But I thought about this: `hold_off` does not really fit into `@time_active` logically. Yes, it is also time-related, but the meaning is different.
Thanks to DecoratorManager, we can now easily decompose decorators instead of extending existing ones just because adding a new decorator used to be a big task.

My proposal:
* mark the `hold_off` parameter in `@time_active` as deprecated, without removing it, only with a warning
* add a new decorator: `@hold_off(float, send_last: False)`

Then everything becomes explicit. If `@time_active` and `@hold_off` are used together, the behavior depends only on the order specified by the user.

Maybe `@hold_off` is not the best name. `@throttle` or another name might be better. Need to think about it.

dmamelin:
Sorry, I misled you.

When `@hold_off` and `@time_active` are used together, the event from `@hold_off` will still be filtered through `@time_active`, because dispatch calls all TriggerHandlerDecorators in order:
```python
decorators = self.get_decorators(TriggerHandlerDecorator)
for dec in decorators:
    if await dec.handle_dispatch(data) is False:
        self.logger.debug("Trigger not active due to %s", dec)
        return
```

My previous message was based on behavior that was never implemented.

When developing DM, I wanted to stay closer to Python behavior, but decided not to because of backward compatibility concerns.
```python
@state_trigger("first_var")
@time_active("first_time")
@state_trigger("second_var")
@time_active("second_time")
```
In both the legacy implementation and DM, `first_var` and `second_var` are filtered by both `first_time` and `second_time`.
The idea was to stay closer to Python and call decorators in order starting from the trigger that caused dispatch.
So for `first_var`, `first_time` and `second_time` would be called. For `second_var`, only `second_time` would be called.
But again, I did not implement this because of backward compatibility issues.
It is not hard to implement, and tests probably would not even need changes, but I am not sure how useful it is.

So, regarding `hold_off_send_last`: in the current architecture, it definitely should check the `@time_active` window.
I can either extract `@hold_off` into a separate decorator, or keep everything inside `@time_active` and finish it in the current PR.

@craigbarratt  what do you think?

craigbarratt:
@dmamelin Good point about the non-Pythonic behavior of consecutive decorators.  That was my mistake in the implementation, so they are cumulative and unordered, as you note.  I don't think it's worth changing.

TBH I don't know why I overloaded `@time_active` with `hold_off`, other than it being a time-related condition.  I don't think it's worth breaking it into a new decorator function, so I'm fine with the current implementation.

There are still the two open questions:
- It `@time_active` includes a time range, and the first trigger occurs (ie, inside the time window), and the extra trigger with `hold_off_send_last=True` falls outside the time active window, should it be suppressed or always occur?  You said the behavior is that it will be suppressed.  I actually prefer it to always happen, since the user code's logic might assume that it always has the latest state value.  If the user really wants to ignore the extra window because it is outside some time range, they can check inside their function.  If we do the opposite, the extra trigger is dropped and the user function can't intervene.  Either way, we should document the behavior in this corner case.
- Is the potential race condition real, and if so, how easy is it to fix?

Thanks for adding this feature!

dmamelin:
- The deferred run never checked `time_spec` ranges to begin with, so it already behaves as you prefer (the latest data always reaches the function). Documented this corner case, including how to keep a range in force via a separate `@time_active`, and added a test pinning the behavior.
- Fixed the race between the `hold_off_send_last` task and a trigger passing right as the window expires: the task now re-checks `_pending_data` after every await, so it can no longer dispatch `None`.
- Docs updated: the decorator manager allows multiple `@time_active` decorators per function.
- Increased `hold_off` in the test from 0.05s to 0.1s: the test relies on two state writes landing within the hold-off window of the first trigger's dispatch, and 50 ms leaves little headroom on a loaded CI runner. Virtual time isn't an option here — the mechanism runs on `time.monotonic()` + `asyncio.sleep`, which is the event loop's own clock.

craigbarratt:
Awesome - thanks for all your work on this!

PR #847, opened 2026-05-19
fix state trigger hold on unrelated state updates

### LINKED ISSUE #845: state_hold in version 2.0 doesn't work in all cases, but still works with legacy_decorators: true
First of all, thanks for all the great work.
I have been using pyscript for years and do pretty much all automation with it. 
I got a trigger if a new movie or episode is played to write it into a database. To avoid too many entries if I accidentally start an episode, I use state_hold to give it a few seconds

@state_trigger("media_player.kodi == 'playing' ", state_hold=10)
def kodi_mark_watched():
  log.debug(f"KODI Media Position:  {state.get('media_player.kodi.media_position')} ") 
...

This worked fine with previous pyscript versions. Now I updated to 2.0. and it doesn´t trigger at all. If i remove the state_hold=10 argument, it triggers. I also tried the  legacy_decorators: true in pyscript.yaml and now it works again. 

What puzzles me is the fact that I use it in other parts of my code, and it is working as expected:
@state_trigger("media_player.kodi == 'playing' and input_boolean.mode_night == 'on' ", state_hold=42)
def kodi_playing():
  pyscript.light_on( device='light.tv_strip',        red=242, green=175, blue=35, bright=30, trans=5)
  pyscript.light_on( device='light.livingroom_bulb', red=242, green=175, blue=35, bright=20, trans=5)

How could I assist to debug that?

dmamelin:
Fixes #845.

The new `state_trigger` treated unrelated state events as a `false` trigger condition. For entities like `media_player`, attribute-only updates can arrive while the watched state remains unchanged, for example `playing`. Those updates could reset `state_hold`, causing the trigger to never fire.

This changes the new `state_trigger` decorator to ignore unrelated state events, matching the legacy trigger behavior. A regression check was added to test_state_trigger for attribute-only updates during state_hold.

### DIFF
--- a/custom_components/pyscript/decorators/state.py
+++ b/custom_components/pyscript/decorators/state.py
@@ -149,11 +149,13 @@ class StateTriggerDecorator(TriggerDecorator, ExpressionDecorator, AutoKwargsDec
             return "None"
         return f"{(now - dt):g} ago"
 
-    async def _check_new_state(self, trig_ok: bool) -> None:
+    async def _check_new_state(
+        self, trig_ok: bool, new_vars: dict[str, Any], func_args: dict[str, Any]
+    ) -> None:
         now = asyncio.get_running_loop().time()
         if _LOGGER.isEnabledFor(logging.DEBUG):
             msg = f"check_new_state: {self}"
-            msg += f"\ntrig_ok: {trig_ok} now {now} func_args: {self.last_func_args} new_vars: {self.last_new_vars}"
+            msg += f"\ntrig_ok: {trig_ok} now {now} func_args: {func_args} new_vars: {new_vars}"
             if self.true_entered_at:
                 msg += f"\ntrue_entered_at: {self.true_entered_at}({(now - self.true_entered_at):g} ago)\n"
             if self.false_entered_at:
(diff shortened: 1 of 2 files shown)


craigbarratt:
Overall looks good.  Is there a potential minor regression with `old_value` now?  Here's the concern from claude:

last_new_vars / last_func_args are overwritten with the unrelated event's data (state.py:276-277) before the branch, then continue skips _check_new_state. So when the hold later fires via _check_state_hold, it dispatches with the latest (unrelated) event's func_args/new_vars rather than the event that started the hold:
```
await self.dispatch(DispatchData(self.last_func_args, trigger_context={"new_vars": self.last_new_vars}))
```
In the old code this never surfaced (the hold was canceled, so it never fired from an unrelated event). The test doesn't distinguish it because value stays "playing" and position is read live from state. But old_value would now reflect the last attribute-only change ("playing" → "playing") rather than the original transition into "playing". Worth a quick check that this "report latest values on hold expiry" semantics matches legacy — if a user keys on old_value, this could differ. Minor, but it's exactly the kind of edge a legacy-parity fix should pin down.

dmamelin:
Great catches, fixed: last_func_args/last_new_vars are now captured only at hold start, matching legacy behavior.

craigbarratt:
Awesome - thanks!

PR #851, opened 2026-05-29
use service response for jupyter kernel

devsnek:
with a compatible client (https://github.com/craigbarratt/hass-pyscript-jupyter/pull/39), just return the ports directly instead of doing a state dance.

at some point in the future the old path can be removed.

### DIFF
--- a/custom_components/pyscript/__init__.py
+++ b/custom_components/pyscript/__init__.py
@@ -354,7 +354,7 @@ async def async_setup_entry(hass: HomeAssistant, config_entry: ConfigEntry) -> b
         DOMAIN, SERVICE_GENERATE_STUBS, generate_stubs_service, supports_response=SupportsResponse.ONLY
     )
 
-    async def jupyter_kernel_start(call: ServiceCall) -> dict | None:
+    async def jupyter_kernel_start(call: ServiceCall) -> dict:
         """Handle Jupyter kernel start call."""
         _LOGGER.debug("service call to jupyter_kernel_start: %s", call.data)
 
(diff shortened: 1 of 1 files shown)


custom_components/pyscript/__init__.py
```diff
@@ -369,14 +369,25 @@ async def jupyter_kernel_start(call: ServiceCall) -> None:
...
 
-        kernel.set_session_cleanup_callback(state_var_remove)
+            def state_var_remove():
+                hass.states.async_remove(call.data["state_var"])
 
-    hass.services.async_register(DOMAIN, SERVICE_JUPYTER_KERNEL_START, jupyter_kernel_start)
+            kernel.set_session_cleanup_callback(state_var_remove)
+
+        # newer kernels use the service response
+        return {'ports': kernel.get_ports()}
```
craigbarratt: Please use double-quotes on "ports"

custom_components/pyscript/__init__.py
```diff
@@ -354,7 +354,7 @@ def write_stubs(path: str) -> dict[str, Any]:
         DOMAIN, SERVICE_GENERATE_STUBS, generate_stubs_service, supports_response=SupportsResponse.ONLY
     )
 
-    async def jupyter_kernel_start(call: ServiceCall) -> None:
+    async def jupyter_kernel_start(call: ServiceCall) -> dict | None:
```
craigbarratt: This only returns a `dict` now, so remove `None`.

craigbarratt:
Thanks for the PR!

PR #852, opened 2026-06-05
Add new webhook handler annotation.

JayNewstrom:
This is the follow up to https://github.com/custom-components/pyscript/pull/843

`@webhook_handler` waits for the function to return, and uses it's return value as the HTTP response. 

After investigation, home assistant already validates unique webhook_id, so I removed that behavior form the original PR.


dmamelin:
Thanks for the work on this. I took a closer look at the patch and I think the feature makes sense, but I am not fully comfortable with the current architecture.

I am working on an alternative implementation based on the same idea, but with a smaller scope and closer integration with the existing decorator flow. It is possible that I will end up with a similar result, but I need a few more days to check this properly.

I will share it here once it is ready, so we can compare both approaches.

craigbarratt:
Will be replaced by #856.

PR #862, opened 2026-07-16
fix(webhook): restore aiohttp request in legacy webhook func_args

diegograssato:
Restores request in the kwargs payload built by the legacy webhook handler.

Why:

User webhook functions may need direct access to aiohttp request
(headers, method, query and raw body).
This is required for HMAC/signature validation and advanced webhook checks.
Aligns behavior with decorators/webhook.py (commit 5612dd4), which already
exposes request to webhook trigger functions.
Compatibility:

Backward compatible for scripts that do not declare request.
Comandos prontos para commit:

cd /home/grassato/workspace/pyscript
git add custom_components/pyscript/webhook.py
git commit -m "fix(webhook): restore aiohttp request in legacy webhook func_args" -m "Restores request in the kwargs payload built by the legacy webhook handler.
Why:

User webhook functions may need direct access to aiohttp request (headers, method, query and raw body).
This is required for HMAC/signature validation and advanced webhook checks.
Aligns behavior with decorators/webhook.py (commit 5612dd4), which already exposes request to webhook trigger functions.
Compatibility:

Backward compatible for scripts that do not declare request."

### DIFF
--- a/custom_components/pyscript/webhook.py
+++ b/custom_components/pyscript/webhook.py
@@ -42,6 +42,7 @@ class Webhook:
         func_args = {
             "trigger_type": "webhook",
             "webhook_id": webhook_id,
+            "request": request,
         }
 
         if "json" in request.headers.get(hdrs.CONTENT_TYPE, ""):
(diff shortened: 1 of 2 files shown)


craigbarratt:
Thanks for the PR.  Can you fix the indicated errors?

The test you added actually tests the new decorator code, not the legacy mode.  I'm actually fine with that since everyone should be using the new decorator code, which already supports the `request` func arg.

PR #863, opened 2026-08-02
Tweaks in type stubs generation

contagon:
Ever since stub generation was built into pyscript, I've been leaning heavily into python typing (using pyrefly as my typechecker currently, previously basedpyright) as I find this has been catching A LOT of the bugs I had in my automations.

There was always a couple of typing oddities the type checkers didn't like in the pyscript stubs. I've been fixing them by hand for a while, but finally took the time to fix them in the generation.

DISCLOSURE: A lot of the code in the PR was LLM generated, but I reviewed it to the best of my ability. Also the PR is all hand-written:)

## Missing entities

I had a number of missing entities in the stubs, namely `zone.home` and a number of template entities were the ones I noticed. This PR now uses the state machine (sorted to get more deterministic output) instead of the entity registry to get ALL entities. From what I can tell, this should include everything the entity registry also did.

## Poor defaults

There was a number of defaults in the stubs that didn't type check correctly. Examples include `delay_secs: int = 0.4`,  `activity: int = None`, and `pipeline : str = 1`. These are now transformed into `delay_secs: float = 0.4` , `arg: int | None = None`, and `pipeline: str | int = 1`. 

## Tests & Type Checkers

Tests have been added for each of these! I'd also be happy to add a section to the docs about setting up a type checker with pyscript. Between stubs giving autocompletion and the type checker making sure I call services correctly (bad types when calling -> type errors) and am only using entities that still exist (deleted entities -> generate new stubs -> type errors anywhere they are used), writing pyscript now feels just like native python code!

I'd recommend giving this a shot before merging, just to make sure it all works! Also happy to tweak whatever

### DIFF
--- a/custom_components/pyscript/stubs/generator.py
+++ b/custom_components/pyscript/stubs/generator.py
@@ -11,7 +11,6 @@ from typing import Any, Literal
 
 from custom_components.pyscript.stubs.pyscript_builtins import StateVal
 from homeassistant.core import HomeAssistant, split_entity_id
-from homeassistant.helpers import entity_registry as er
 from homeassistant.helpers.service import async_get_all_descriptions
 
 _LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


craigbarratt:
Thanks for the PR!

PR #872, opened 2026-09-11
fix: check dest_path when filtering watchdog reload events

### LINKED ISSUE #871: No automatic reload after editing files
Today I updated HA from core 2026.8.3 to 2026.9.1

Before the HA update, a pyscript that was modified, automatically reloaded,

Now I have to manually reload (the whole integration or calling the pyscript.reload) to activate the script version

dmamelin:
Fixes #871

Editors that save atomically (temp file + `os.replace`/`rename` — e.g. `Blueprint Studio`) surface as a `FileMovedEvent` where `src_path` is the temp file name and `dest_path` is the real script path. `check_event` only checked `src_path`, so these saves were silently ignored and the reload never fired.

Now both `src_path` and `dest_path` are checked against the allowed suffixes.

### Testing

`check_event` currently has no dedicated tests - the watchdog wiring itself is mocked out everywhere it's touched. Given the fix is a one-line, low-risk check (also inspect `dest_path` for moved events), I kept this PR minimal rather than doing an unrelated refactor to make `check_event` independently testable. Happy to extract it to a module-level function and add unit tests if you'd like coverage for this specific case.


craigbarratt:
Thank for the PR - looks good.

